Issue #23, opened 2022-01-31
Add documentation on creating a Swift Package


Jomy10:
Hi, I'm the one who asked about using Rust in Swift on Reddit.

I've made a Gist that follows your steps, but slightly modified to set up a Swift project with Swift Package Manager ([https://gist.github.com/Jomy10/a4873dd43942ed1bf54d387dbc888795](https://gist.github.com/Jomy10/a4873dd43942ed1bf54d387dbc888795)). 

I've tried to add dependencies to `Package.swift`, but I don't know how to include them in the compiling process.  Swiftc documentation seems to be rather scarce.

Here's an example of the `Package.swift` file:

```swift
// swift-tools-version:5.5
import PackageDescription

let package = Package(
    name: "swift",
    dependencies: [
        .package(url: "/path/to/HelloSwift", .branch("master"))
    ],
    targets: [
        .target(
            name: "swift",
            dependencies: [
                .product(name: "HelloSwift", package: "HelloSwift")
            ]),
        .testTarget(
            name: "swiftTests",
            dependencies: ["swift"]),
    ]
)
```

Folder structure is also in the gist.

I think it would be great if we could also use Swift Package Manager in the Swift project using the Rust library. Or maybe import the Rust library as a package in `Package.Swift`.

chinedufn:
Hey!

If I'm understanding, you're saying that you'd like to be able to have the Rust side of things get packaged up into a Swift package that you can then import from your Swift project? Or are you saying something else?

I'll read up on Swift packages since I don't know much about them at the moment.

Would using Swift packages be strictly better to what we're currently doing, or are there any downsides?

Thanks for opening this issue!

chinedufn:
Alright I read up on Swift packages on modules and have a basic understanding.

It seems like if I had to find the closest Rust ecosystem version of them..:

Swift Package == Cargo crate
Swift Module == Rust module

I know they aren't the same.. I'm just going for a rough understanding of what these Swift code organization tools are.

---

Ok so I'd sum up my current thoughts as:

- I don't yet understand what your end goal is and would love an explanation / some context around your use case so that I can better understand why you want to be able to generate a Swift package.

Jomy10:
I do think that compiling to a Swift package could be a huge plus. Right now we can compile Swift code using `swiftc` and linking the rust library. This is great for cli's. However, if we could either compile the Rust code to a Swift library, or compile the Swift project that includes the Rust code to a Swift library, I think this could be a huge benefit.

[...]

PR #30, opened 2022-02-16
Swift package book chapter

Jomy10:
I have added the chapter on creating a Swift package to the book (#23).

Feel free to suggest impovements ofcourse.


book/src/building/swift-package/README.md
```diff
@@ -0,0 +1,246 @@
+# Bundling rust code as a Swift package
+
+It is possible to bundle a rust library into a Swift package. This solution only works when targetting Apple platforms, though.
+
+## Project setup
+
+```bash
+mkdir swift_rust_project && cd $_
```
chinedufn, line 8: Naming suggestion

`mkdir rust_swift_package && cd rust_swift_package`

book/src/building/swift-package/README.md
```diff
@@ -0,0 +1,246 @@
+# Bundling rust code as a Swift package
+
+It is possible to bundle a rust library into a Swift package. This solution only works when targetting Apple platforms, though.
```
chinedufn, line 3:
```
In this chapter we'll walk through bundling your Rust library into a Swift Package.

Note that Swift Packages that contain native libraries only work on Apple hardware. 

You should avoid bundling your Rust code into a Swift Package if you plan to target Linux, Windows or any other non-Apple target.
```

book/src/building/swift-package/README.md
```diff
@@ -0,0 +1,246 @@
...
+    header "SwiftBridgeCore.h"
+    export *
+}
+```
+
+Now it is time to build the xcframework:
+
+```bash
+xcodebuild -create-xcframework \
+    -library simulator/librust.a \
```
chinedufn, line 112: I don't think these lib names will work now that you've renamed them

book/src/building/swift-package/README.md
```diff
@@ -0,0 +1,246 @@
...
+│   └── MySwiftPackage
+│       └── SwiftBridgeCore.swift
+│       └── my_rust_lib.swift
+├── MyRustLib.b.xcframework
+└── Package.swift
+```
+
+Add the framework as a binary target
+```swift
+// Package.swift
```
chinedufn, line 139: `MySwiftPackage/Package.swift`

book/src/building/swift-package/README.md
```diff
@@ -0,0 +1,246 @@
...
+```
+
+or 
+
+```swift
+// Sources/MySwiftPackage/my_rust_lib.swift
+// ...
+// Change return type to `String` and call `.toString`
+public func hello_rust() -> String {
+    RustString(ptr: __swift_bridge__$hello_rust()).toString()
```
chinedufn, line 201: By the way, in general we shouldn't recommend that people hand edited the generated code.

book/src/building/swift-package/README.md
```diff
@@ -0,0 +1,246 @@
...
+
+// ...
+```
+
+```swift
+// Sources/MySwiftPackage/my_rust_lib.swift
+import MyRustLib
+```
+
+We can choose to either make the functions and classes in `SwiftBridgeCore.swift` public, or to convert the return types of our functions to types recognized by swift, so in this case, do either of the following:
```
chinedufn, line 177: Oh ok, so the issue here is that the `toString()` function and other functions are private?

You can make `toString` public here:

https://github.com/chinedufn/swift-bridge/blob/771fc5cae52b0206b43135ee3adbc286f6cd64bd/src/std_bridge/string.swift#L27-L34

Then I think we can get rid of this suggestion?

book/src/building/swift-package/README.md
```diff
@@ -0,0 +1,246 @@
...
+
+## Using the Swift Package
+
+We now have a Swift Package which we can include in other projects using the Swift Package Manager.
+
+### Example: MacOS executable
+Here is an example of an executable project located in `swift_rust_project/testPackage`.
+
+```swift
+// Package.swift
```
chinedufn, line 213: `testPackage/Package.swift` I think?

book/src/building/swift-package/README.md
```diff
@@ -0,0 +1,246 @@
...
+            ])
+    ]
+)
+```
+
+```swift
+import MySwiftPackage
+
+print(hello_rust().toString())
+// Or, if you converted the return type of the function to a `String`:
```
chinedufn, line 233: Can remove this second approach after making `toString()` public

book/src/building/swift-package/README.md
```diff
@@ -0,0 +1,246 @@
...
+```
+
+**Output**
+```
+Hello Rust!
+```
+
+### Example: iOS app
+
+To add the package to an iOS app in XCode, go to the target's general panel, click the `+` button in the `Frameworks, Libraries, and Embedded Content` section. Then, click `Add Other` and choose `Add Package Dependency`.
```
chinedufn, line 244: Awesome!!

chinedufn (changes requested):
@Jomy10 really, really awesome stuff here!

Thanks for putting this all together. I like how you used a project based approach to really spell out to the reader how to make the package.

In the future we'll automate a lot of this of course.. (perhaps behind `swift-bridge-cli`).. but in the meantime this serves as a great reference for how to hook all the pieces together.

I left a few small bits of feedback, but overall this is looking solid.

Thanks for your work!

book/src/building/swift-package/README.md
```diff
@@ -0,0 +1,246 @@
+# Bundling rust code as a Swift package
```
chinedufn, line 1: Need to add this file to `book/src/SUMMARY.md`

Also should capitalize `Rust` across the file.

chinedufn (commented):
Left a couple more tiny notes

src/std_bridge/string.swift
```diff
@@ -25,7 +25,7 @@ class SwiftString {
 }
 
 extension RustString {
-    func toString() -> String {
+    public func toString() -> String {
```
Jomy10, line 28: By the way, I'm not sure which other methods should be declared public as well

Jomy10:
All comments should be addressed now in my changes.

book/src/building/swift-package/README.md
```diff
@@ -0,0 +1,221 @@
...
+            dependencies: ["MyRustLib"]),
+    ]
+)
+```
+
+We will need to import our rust library in the `.swift` files:
+
+```swift
+// MySwiftPackage/Sources/MySwiftPackage/SwiftBridgeCore.swift
+import MyRustLib
```
chinedufn, line 171: Why does `SwiftBridgeCore` need to import anything? We shouldn't touch this core library file.

book/src/building/swift-package/README.md
```diff
@@ -0,0 +1,221 @@
...
+
+// ...
+```
+
+```swift
+// MySwiftPackage/Sources/MySwiftPackage/my_rust_lib.swift
+import MyRustLib
```
chinedufn, line 178: Is it possible to not touch this generated file, or is this necessary? 
I'd like to avoid touching this file if possible.

i.e. can there just me another file that imports this and imports `MyRustLib` or something like that..?

book/src/building/swift-package/README.md
```diff
@@ -0,0 +1,221 @@
...
+            name: "testPackage",
+            dependencies: [
+                .product(name: "package", package: "package")
+            ])
+    ]
+)
+```
+
+```swift
+testPackage/Sources/testPackage/main.swift
```
chinedufn, line 206: Missing `//`

src/std_bridge/string.swift
chinedufn: Right for sure - no worries we can handle that over time.

book/src/building/swift-package/README.md
```diff
@@ -0,0 +1,221 @@
...
+            path: "MyRustLib.xcframework"
+        ),
+        .target(
+            name: "MySwiftPackge",
+            dependencies: ["MyRustLib"]),
+    ]
+)
+```
+
+We will need to import our rust library in the `.swift` files:
```
chinedufn, line 167: Why? Is there an alternative approach we can take instead of modifying our generated files?

book/src/building/swift-package/README.md
```diff
@@ -0,0 +1,221 @@
...
+    -library simulator/libmy_rust_lib.a \
+    -headers include \
+    -library ios/libmy_rust_lib.a \
+    -headers include \
+    -library macos/libmy_rust_lib.a \
+    -headers include \
+    -output MyRustLib.xcframework
+```
+
+*The order of the `library` tags is important*
```
chinedufn, line 125: Why?

book/src/building/swift-package/README.md
```diff
@@ -0,0 +1,221 @@
...
+## Project setup
+
+```bash
+mkdir rust_swift_package && cd rust_swift_package
+```
+
+### Rust project setup
+
+```bash
+cargo new my_rust_lib --lib
```
chinedufn, line 18: `my-rust-lib`

book/src/building/swift-package/README.md
```diff
@@ -0,0 +1,221 @@
...
+In this chapter we'll walk through bundling your Rust library into a Swift Package.
+
+Note that Swift Packages that contain native libraries only work on Apple hardware. 
+
+You should avoid bundling your Rust code into a Swift Package if you plan to target Linux, Windows or any other non-Apple target.
+
+## Project setup
+
+```bash
+mkdir rust_swift_package && cd rust_swift_package
```
chinedufn, line 12: `rust-swift-package`

chinedufn (commented):
Awesome stuff.

One more piece of real feedback around how we're modifying the generated code.. and some tiny cleanup feedback.

We're close to the finish line! Once we sort out the part where we're modifying generated code we should be good to go!

Awesome work!

book/src/building/swift-package/README.md
Jomy10: I think you're right here. I didn't give this a second thought after testing, but the core library indeed doesn't have to import this.

book/src/building/swift-package/README.md
Jomy10: I'm not sure that this is possible. This file needs access to `__swift_bridge__$hello_rust()`, which is a function defined in `MyRustLib`.

book/src/building/swift-package/README.md
Jomy10: I haven't been able to find any documentation on this, but when I put `-library macos/libmy_rust_lib.a` on top, it doesn't work, it only works in the order simulator - iOS - macOS. 

If I can find any documentation on this, I'll add it.

Jomy10:
New changes should address your comments. I've left a few comments regarding the questions.

book/src/building/swift-package/README.md
chinedufn: Ok great. This should do:

*The order of the `library` flags is important, but we don't currently know why*

Jomy10: Awesome, I have changed it

chinedufn:
Ok this looks great.

Last but not least can you follow your own instructions locally from scratch and verify that everything works.

Once you confirm that these docs can be followed from scratch we're good to merge!

book/src/building/swift-package/README.md
```diff
@@ -0,0 +1,214 @@
...
+    -output MyRustLib.xcframework
+```
+
+*The order of the `library` tags is important, but we don't currently know why*
+
+## Creating the Swift package
+
+Go back to the root of the project and create a new Swift package, copy the xcframework and generated swift files to it:
+
+```
```
chinedufn, line 131: Same here, can we add the bash for making this directory and copying the files to it like in my comment above?

book/src/building/swift-package/README.md
chinedufn: I think you'd need to handle this in the `Package.swift`.

I think this is explained here https://stackoverflow.com/questions/65220359/add-package-dependency-for-a-binary-target-with-swift-package-manager/65255303#65255303

Where basically you create a wrapper target that depends on the binary framework and the Swift code. Then we'd have the library product expose that wrapper target.

Note that I've never actually made a Swift package so I'm just going off of Google.

I'm also totally cool with you just leaving a TODO comment in the documentation for now and we can figure this out later so that this PR can land.

Thanks!

book/src/building/swift-package/README.md
```diff
@@ -0,0 +1,214 @@
...
+cargo build --target aarch64-apple-ios
+cargo build --target x86_64-apple-ios
+```
+
+## Creating the XCFramework
+
+Go back to the root of the project and make a new directory.
+
+```bash
+mkdir MyFramework && cd $_
```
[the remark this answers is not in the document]
Jomy10: Good idea! I'll change it this evening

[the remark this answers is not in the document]
chinedufn: Awesome! Yeah ideally anyone going through our documentation can just copy and paste things. 

For example instead of only saying "create the file foo" we say something like 

Create the file foo.bar

`touch foo.bar`

---

Ok awesome. Thanks a lot for cleaning all of this up. Your work on Swift Package support is going to be really valuable for being able to expose Rust code to Swift.

Thanks again!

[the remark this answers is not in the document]
Jomy10: I've learned a lot about Swift Packages from doing this, so it was a pleasure to figure this out.

book/src/building/swift-package/README.md
Jomy10: I'm not sure that would work, as the swift package also depends on the binary framework. I can look into it some other day, though.

book/src/building/swift-package/README.md
```diff
@@ -0,0 +1,258 @@
...
+import MyRustLib
+```
+
+<!--TODO: always make public, or maybe just have an option for making functions public in build.rs?-->
+```swift
+// MySwiftPackage/Sources/MySwiftPackage/my_rust_lib.swift
+import MyRustLib
+
+// Also make the functions here public
+public hello_rust() -> RustString {
```
Jomy10, line 211: I have asked the user to make the functions public for now. I don't know if we should make SwiftBridge automatically make these public, though. Maybe we have a config option or env variable to opt in to public functions?

book/src/building/swift-package/README.md
```diff
@@ -201,9 +201,14 @@ We will need to import our rust library in `my_rust_lib.swift` and `SwiftBridgeC
 import MyRustLib
 ```
 
+<!--TODO: always make public, or maybe just have an option for making functions public in build.rs?-->
 ```swift
 // MySwiftPackage/Sources/MySwiftPackage/my_rust_lib.swift
 import MyRustLib
+
+// Also make the functions here public
+public hello_rust() -> RustString {
```
chinedufn, line 211: You can make this `public func`

https://github.com/chinedufn/swift-bridge/blob/72c2fa5f2214e0845d9b3fa29b56ffbf94a97c68/crates/swift-bridge-ir/src/codegen/generate_swift.rs#L409

And then it will work

Jomy10:
Alright, I have tried this from scratch, and it is working, apart from some typos, which I have fixed.

Shell commands have been added for copying and creating files, etc.

I have left a comment about making functions in generated files public.

chinedufn (approved):
Left one more thing based on the new changes.

Once you confirm that following these new docs actually work, and tests are passing, we're good to merge!

book/src/building/swift-package/README.md
chinedufn: I guess it's technically a bit cleaner if we're explicitly importing the Rust that we're using so.. I suppose this isn't the worst thing.

I just don't like the idea of manually editing a generated file.

But.. all of this is temporary until we automate this.. So.. works for now!

book/src/building/swift-package/README.md
chinedufn: I guess it's technically a bit cleaner if we're explicitly importing the Rust that we're using so.. I suppose this isn't the worst thing.

I just don't like the idea of manually editing a generated file.

But.. all of this is temporary until we automate this.. So.. works for now!

crates/swift-bridge-ir/src/codegen/generate_swift.rs
```diff
@@ -401,7 +401,7 @@ fn gen_func_swift_calls_rust(
     let swift_class_func_name = if function.is_swift_initializer {
         "convenience init".to_string()
     } else {
-        format!("func {}", fn_name.as_str())
+        format!("public func {}", fn_name.as_str())
```
chinedufn, line 404: `cargo test -p swift-bridge-ir`

A bunch of tests will probably be failing. You just need to add the word `public to them and they should pass!

Jomy10:
All tests should be passing now

chinedufn:
Hmm it looks like there are 5 failing integration tests https://github.com/chinedufn/swift-bridge/runs/5237505800?check_suite_focus=true

You can run them in the command line using `./test-integration.sh`

You can open the integration tests in Xcode using `open SwiftRustIntegrationTestRunner`

Jomy10:
I see. I'll fix it later!

Jomy10:
Allright, everything should be passing right now!

chinedufn:
Thanks a lot for your first contribution!

PR #36, opened 2022-03-08
Generate Swift Package

Jomy10:
I have added support for generating Swift Packages.

The `generate_package` function inside of `package.rs` creates a package containing the xcframework needed. 

The `GeneratePackageConfig` has 4 fields:
- `bridge_dir`: the directory containing the generated bridge files
- `paths`: the path to the compiled rust library and its platform. I have made the `ApplePlatform` enum follow the naming conventions of Apple. Let me know if you would rather have Rust naming conventions for this.
- `out_dir`: the output directory for th swift package
- `package_name`: the name for the Swift Package

I have added a test that generates a Swift Pacakage and tests that it executes and the output is correct.


chinedufn:
I'm so excited about your work on this!

I'll review this today or tomorrow, but here are a few quick notes from me quickly glancing at the diff:

- Tests are failing. According to the logs my guess is there's a typo somewhere: `SwiftBirdgeCore`

- Steps in the `Swift Package` chapter that we no longer need to do by hand should be replaced by a call to  `swift_bridge_build::generate_package`.
  - Similarly, when we introduce `swift-bridge-cli`'s generate package subcommand we'll want to add that to the book

- We shouldn't manually add the `import` at the top of the generated files anymore. We should instead
  1. Copy the bridge dir into a temp dir ( i.e. with https://github.com/Stebalien/tempfile ) so that our build process doesn't mutate the user's files. All of our building should happen in that temporary directory (so that there is no junk in the user's out dir if the program were to crash) and when we're done we can move whatever we need over to the out dir.
  2. Prepend the import line to the necessary files
  3. Use that temp dir instead of the bridge dir for the rest of the process

I'll review more closely later and leave comments on the PR. This is just some quick feedback.

This is really exciting work! Awesome stuff.

Jomy10:
Awesome!

The tests succeeded for me locally, so I will look into that.

I'll also update the book chapter to explain the new steps.

Also a last note, all error handling currently just exits the process. I thought this was appropriate in this situation, but correct me if I'm wrong.

I will work on the issues this week.

chinedufn:
That error handling approach sounds fine for now!

Jomy10:
I have updated the book chapter. Let me know what you think of the current approach (since the cli is not updated yet).

Jomy10:
The test should succeed now, the only problem is that the GitHub action doesn't have the `x86_64-apple-darwin`target installed. I'm also not sure how to add this to to the action.

Cargo.toml
```diff
@@ -33,9 +33,10 @@ members = [
   "crates/swift-bridge-cli",
   "crates/swift-bridge-ir",
   "crates/swift-bridge-macro",
-  "crates/swift-integration-tests",
+
+  "crates/swift-integration-tests","crates/swift-bridge-build/tests/sample-project",
```
chinedufn, line 37: Instead of `sample-project` we should name the crate `test-swift-packages`

book/src/building/swift-packages/README.md
```diff
@@ -53,159 +53,67 @@ Add a new `build.rs` file (`touch build.rs`):
...
-│   └── libmy_rust_lib.a
-├── macos
-│   └── libmy_rust_lib.a
-└── simulator
-    └── libmy_rust_lib.a
-```
+Create a new folder called *generated* `mkdir generated`.
 
-Edit `include/module.modulemap`:
+Create a new rust project called "post-build":
```
chinedufn, line 73: This isn't a common pattern. So, instead of suggesting that people make a build crate we can just tell them that the generate package command exists and call it a day.

Something along the lines of:

---

After parsing your bridge modules and writing the bridging code to disk, you can take that bridging code and package it into a Swift Package.

First, build your Rust library for the desired platforms:

```bash
export SWIFT_BRIDGE_OUT_DIR="$(pwd)/generated"
cargo build --target x86_64-apple-darwin
cargo build --target aarch64-apple-ios
cargo build --target x86_64-apple-ios
```

Then use the `API` or `CLI` to package the bridging code and the Rust libraries into a Swift Package.

## API

```rust
use std::path::Path;
use std::collections::HashMap;
use swift_bridge_build::{GeneratePackageConfig, ApplePlatform};
fn main() {
    swift_bridge_build::generate_package(GeneratePackageConfig {
        bridge_dir: &Path::new("./generated"),
        paths: HashMap::from([
            (ApplePlatform::iOS, &"target/x86_64-apple-ios/debug/libmy_rust_lib.a" as _),
            (ApplePlatform::Simulator, &"target/aarch64-apple-ios/debug/libmy_rust_lib.a" as _),
            (ApplePlatform::macOS, &"target/x86_64-apple-darwin/debug/libmy_rust_lib.a" as _),
        ]),
        out_dir: &"MySwiftPackage",
        package_name: "MySwiftPackage"
    });
}
```

## CLI

Not yet implemented

crates/swift-bridge-build/src/package.rs
```diff
@@ -0,0 +1,287 @@
...
+
+use std::collections::HashMap;
+use std::fs;
+use std::fs::OpenOptions;
+use std::io::Write;
+use std::path::Path;
+use std::process::Command;
+
+/// Config for generating Swift packages
+pub struct GeneratePackageConfig<'a> {
```
chinedufn, line 11: Minor note:

How do you feel about using the word `create` instead of generate?

i.e. `CreatePackageConfig` .. `fn create_package` etc.

I'm thinking that generate sounds like we're doing a bunch of code generation, when really we're mainly taking things that already exist and packaging them together.

crates/swift-bridge-build/src/package.rs
```diff
@@ -0,0 +1,287 @@
...
+	/// The directory containing the generated bridges
+	pub bridge_dir: &'a dyn AsRef<Path>,
+    /// Path per platform. e.g. `(ApplePlatform::iOS, "target/aarch64-apple-ios/debug/libmy_rust_lib.a")`
```
chinedufn, line 14: Run `cargo fmt` on the repository

crates/swift-bridge-build/src/package.rs
```diff
@@ -0,0 +1,287 @@
...
+		Self {
+			bridge_dir,
+			paths,
+			out_dir,
+			package_name
+		}
+	}
+}
+
+#[allow(non_camel_case_types)]
```
chinedufn, line 34: More of a personal preference, but would prefer using Rust style here. If you strongly disagree that's fine too. No big deal.

crates/swift-bridge-build/src/package.rs
```diff
@@ -0,0 +1,287 @@
...
+	/// no official Rust target for this platform
+	watchOSSimulator,
+	/// no official Rust target for this platform
+	carPlayOS,
+	/// no official Rust target for this platform
+	carPlayOSSimulator,
+}
+
+impl ApplePlatform {
+	/// The directory name inside of the xcframework for the specified platform.
```
chinedufn, line 63: Awesome

crates/swift-bridge-build/src/package.rs
```diff
@@ -0,0 +1,287 @@
...
+		}
+	}
+	
+}
+
+/// Generates an xcframework embedded in a Swift Package from the Rust project.
+///
+/// - Also see the [relevant book chapter](https://chinedufn.github.io/swift-bridge/building/swift-packages/index.html)
+/// 
+/// # Parameters
```
chinedufn, line 84: No need for this. You can move any relevant documentation to the `GeneratePackageConfig` type.

If someone is viewing this they can fire up `rustdoc` and read the documentation for `GeneratePackageConfig` to better understand what it is.

crates/swift-bridge-build/src/package.rs
```diff
@@ -0,0 +1,287 @@
...
+/// - `config`: The config for generating the swift package, contains the directory
+///    containing te bridges, the paths to the libraries per platform and the output directory
+pub fn generate_package(config: GeneratePackageConfig) {
+	// Create output directory //
+	let output_dir: &Path = config.out_dir.as_ref();
```
chinedufn, line 89: Don't think the type annotation is necessary, but no big deal.

crates/swift-bridge-build/src/package.rs
```diff
@@ -0,0 +1,287 @@
...
+	gen_xcframework(&output_dir, &config);
+	
+	// Generate Swift Package //
+	gen_package(&output_dir, &config);
+}
+
+/// Generates the XCFramework
+fn gen_xcframework(output_dir: &Path, config: &GeneratePackageConfig) {
+	// Create directories
+	let tmp_framework_path = std::env::temp_dir().join("swiftbridge._tmp_framework");
```
chinedufn, line 104: Do we need to worry about the fact that `std::env::temp_dir` isn't secure?

I don't have any thoughts off the top of my head, just checking.

crates/swift-bridge-build/tests/package-test.rs
```diff
@@ -0,0 +1,45 @@
...
+use std::io::ErrorKind;
+use std::path::Path;
+use std::process::Command;
+
+use swift_bridge_build::{generate_package, GeneratePackageConfig};
+use swift_bridge_build::ApplePlatform as Platform;
+
+#[test]
+fn gen_package() {
+    // Build Rust
```
chinedufn, line 12: We should delete the `generated` dir at the start of the test so that we're 100% sure that we're starting from a clean state.

crates/swift-bridge-build/tests/sample-project/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "sample_project"
+version = "0.1.0"
+edition = "2021"
+
```
chinedufn, line 5: `publish = false`

crates/swift-bridge-build/tests/sample-project/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "sample_project"
```
chinedufn, line 2: `sample-project`

crates/swift-bridge-build/tests/sample-project/build.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+export SWIFT_BRIDGE_OUT_DIR="$(pwd)/generated"
```
chinedufn, line 3: Need to use the same `THISDIR` trick that we use elsewhere. Otherwise if you run this from another directory you'll run into trouble.

i.e. if you did `./crates/swift-bridge-build/tests/sample-project/build.sh` you would end up creating `./generated` instead of `./crates/swift-bridge-build/tests/sample-project/generated`

crates/swift-bridge-build/tests/sample-project/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+#[swift_bridge::bridge]
+mod ffi {
+    extern "Rust" {
+        fn hello_rust() -> String;
+    }
+}
+
+fn hello_rust() -> String {
+    String::from("Hello Rust!")
```
chinedufn, line 9: `"Hello, From Rust!"` just to make it crystal clear that the message came from Rust (as opposed to someone thinking that `Swift` is saying hi to `Rust`.

chinedufn (changes requested):
Awesome work!!

Just some minor feedback.

crates/swift-bridge-build/src/package.rs
Jomy10: create sounds good!

Jomy10:
Alright, I have addressed all your comments. 

The test still isn't passing, because the GH action doesn't have the macOS target installed. Do you have any thoughts?

Cargo.toml
chinedufn: Oh sorry I meant to delete this message! `sample-project` was a fine name.

Whatever you want to go with is fine here.

chinedufn: Oh no worry ignore this. This actually is the test. I think?

Jomy10: This is the Cargo project that is used in the test.

chinedufn:
> The test still isn't passing, because the GH action doesn't have the macOS target installed. Do you have any thoughts?

I'm not sure if we can easily compile targetting `macOS` from a linux machine. The `swift-bridge-build` tests currently run on Linux

https://github.com/chinedufn/swift-bridge/blob/87475eca8889e2d6ce39095868643082b411b58d/.github/workflows/test.yml#L27

---

So, we just need to run this test on `macOS`.

So maybe something like:

- Move `crates/swift-bridge-build/tests/test-swift-packages` -> `SwiftRustIntegrationTestRunner/swift-package-rust-library-fixture`

- In the build script for the `SwiftRustIntegrationTestRunner`, build that Rust library and package it into a swift package https://github.com/chinedufn/swift-bridge/blob/87475eca8889e2d6ce39095868643082b411b58d/SwiftRustIntegrationTestRunner/build-rust.sh#L29

- Make the `SwiftRustIntegrationTestRunner` depend on that Swift package

- Add a new `SwiftPackageTests.swift` file to https://github.com/chinedufn/swift-bridge/tree/87475eca8889e2d6ce39095868643082b411b58d/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests in Xcode

- In that file add a test where we import our swift package and call a function on it. For example, you can make the Rust library expose `fn add_two_numbers(val1: u8, val2: u8) -> u16` and then call it from the Swift test and verify that the results are correct

Let me know if that makes sense!

Jomy10:
Can I put the code to generate the Swift Package inside of `test-integration.sh`? Since the XCode project depends on the generated Swift Package, it has to be present when the `xcodebuild` command is executed, otherwise it will not start building and thus will not execute the `build-rust.sh` script.

chinedufn:
Does this approach work? https://forums.swift.org/t/how-to-add-local-swift-package-as-dependency/26457/8

image

Jomy10:
I did some digging, and couldn't find anything that would allow for this. So I'm testing the package like I was doing before, but moved to the new location. The tests should now succeed if all goes well.

test-integration.sh
```diff
@@ -13,6 +13,27 @@ cd $ROOT_DIR
...
+# Build Rust
+mkdir swift-package-rust-library-fixture/generated
+
+./swift-package-rust-library-fixture/build.sh
+
+# Create Swift Package
+cargo run -p integration-test-create-swift-package
+
+# Test Swift Package
+echo "TESTING SWIFT PACKAGE"
```
chinedufn, line 31: Let's remove this echo

chinedufn (approved):
Awesome work! One small note then looks good to me.

chinedufn:
Can you rebase everything into a single commit and then I'll merge? Thanks!

Jomy10:
Alright, I’m not at my computer right now. I’ll rebase tomorrow!

chinedufn:
No worries, I'll just use the "squash and merge" button which should still keep you at the author of the commit.

Thanks for all of your work here! And for getting this tested!

PR #38, opened 2022-03-11
Add create-package command to CLI

Jomy10:
I have added the cli and the `create-package` subcommand

Let me know if you think my approach is good or not.

### DIFF
--- a/crates/swift-bridge-build/src/package.rs
+++ b/crates/swift-bridge-build/src/package.rs
@@ -37,7 +37,7 @@ impl<'a> CreatePackageConfig<'a> {
     }
 }
 
-#[derive(Debug, Eq, PartialEq, Hash)]
+#[derive(Debug, Eq, PartialEq, Hash, Copy, Clone)]
 /// Currently supported platforms for genereting Swift Packages.
 pub enum ApplePlatform {
     /// `aarch64-apple-ios`
@@ -66,12 +66,12 @@ pub enum ApplePlatform {
 
 impl ApplePlatform {
     /// The directory name inside of the xcframework for the specified platform.
-    fn dir_name(&self) -> &str {
+    pub fn dir_name(&self) -> &str {
         match self {
             ApplePlatform::IOS => "ios",
             ApplePlatform::Simulator => "simulator",
             ApplePlatform::MacOS => "macos",
-            ApplePlatform::MacCatalyst => "macos-mac-catalyst",
+            ApplePlatform::MacCatalyst => "mac-catalyst",
             ApplePlatform::TvOS => "tvos",
             ApplePlatform::WatchOS => "watchos",
             ApplePlatform::WatchOSSimulator => "watchos-simulator",
@@ -79,6 +79,19 @@ impl ApplePlatform {
             ApplePlatform::CarPlayOSSimulator => "carplay-simulator",
         }
     }
+    
+    /// Array containing all `ApplePlatform` variants
+    pub const ALL: &'static [Self] = &[
+        ApplePlatform::IOS,
+        ApplePlatform::Simulator,
+        ApplePlatform::MacOS,
+        ApplePlatform::MacCatalyst,
+        ApplePlatform::TvOS,
+        ApplePlatform::WatchOS,
+        ApplePlatform::WatchOSSimulator,
+        ApplePlatform::CarPlayOS,
+        ApplePlatform::CarPlayOSSimulator
+    ];
 }
 
 /// Generates an xcframework embedded in a Swift Package from the Rust project.
--- a/crates/swift-bridge-cli/src/clap_exec.rs
+++ b/crates/swift-bridge-cli/src/clap_exec.rs
@@ -1,0 +1,39 @@
+use std::collections::HashMap;
+use std::path::{Path, PathBuf};
+use clap::ArgMatches;
+use swift_bridge_build::{ApplePlatform, create_package, CreatePackageConfig};
+
+/// Executes the correct function depending on the cli input
+pub fn handle_matches(matches: ArgMatches) {
+    match matches.subcommand_name() {
+        Some("create-package") => handle_create_package(matches.subcommand_matches("create-package").unwrap()),
+        _ => unreachable!("No subcommand or unknown subcommand given") // Shouldn't happen
+    }
+}
+
+/// Executes the `create-package` command
+fn handle_create_package(matches: &ArgMatches) {
+    let bridges_dir = matches.value_of("bridges-dir").unwrap(); // required
+    let out_dir = matches.value_of("out-dir").map(|p| Path::new(p)).unwrap(); // required
+    let name = matches.value_of("name").unwrap(); // required
+    
+    let mut config = CreatePackageConfig {
+        bridge_dir: &Path::new(bridges_dir),
+        paths: HashMap::new(),
+        out_dir: &out_dir,
+        package_name: name
+    };
+    
+    let platforms = ApplePlatform::ALL;
+    let mut paths: Vec<(ApplePlatform, PathBuf)> = Vec::new();
+    for platform in platforms {
+        if let Some(path) = matches.value_of(platform.dir_name()) {
+            paths.push((*platform, PathBuf::from(path)));
+        }
+    }
+    paths.iter().for_each(|(platform, path)| {
+        config.paths.insert(*platform, path);
+    });
+    
+    create_package(config);
+}
--- a/crates/swift-bridge-cli/src/lib.rs
+++ b/crates/swift-bridge-cli/src/lib.rs
@@ -1,1 +1,18 @@
+#![deny(missing_docs)]
+//! # The swift-bridge CLI.
+//!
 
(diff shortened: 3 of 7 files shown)


crates/swift-bridge-build/src/package.rs
```diff
@@ -37,7 +37,7 @@ impl<'a> CreatePackageConfig<'a> {
     }
 }
 
-#[derive(Debug, Eq, PartialEq, Hash)]
+#[derive(Debug, Eq, PartialEq, Hash, Clone)]
```
chinedufn, line 40: `Copy, Clone`

crates/swift-bridge-cli/src/lib.rs
```diff
@@ -1 +1,7 @@
+mod clap_app;
```
chinedufn, line 1: `#![deny(missing_docs)]` at the top of the file

crates/swift-bridge-cli/src/clap_exec.rs
```diff
@@ -0,0 +1,47 @@
...
+    ];
+    platforms.iter().for_each(|(platform, path)| {
+        if let Some(path) = path {
+            config.paths.insert(platform.clone(), path);
+        }
+    });
+    
+    create_package(config);
+    
+    println!("Package written out to {}", out_dir.display());
```
chinedufn, line 46: Our commands should be silent by default. Mind removing this?

crates/swift-bridge-cli/src/main.rs
```diff
@@ -1 +1,5 @@
-fn main() {}
+use swift_bridge_cli::app::{cli, handle_matches};
```
chinedufn, line 1: The book chapter on Swift packages should mention the `swift-bridge` CLI command

i.e. with a little bit of text and an example like:

```
cargo install -f swift-bridge-cli
swift-bridge create-package ...
```

Similar to what we do with the API section of the `Swift Package` chapter

crates/swift-bridge-cli/src/lib.rs
```diff
@@ -1 +1,7 @@
...
+mod clap_exec;
 
+pub mod app {
```
chinedufn, line 4: Can we just export a single `run()` function (or some other name) that the `main.rs` calls? 

And in that function we can call `handle_matches(cli().get_matches());`?

crates/swift-bridge-cli/src/clap_exec.rs
```diff
@@ -0,0 +1,47 @@
...
+        package_name: name
+    };
+    let platforms = vec![
+        (Platform::IOS, ios_path), (Platform::Simulator, simulator_path), (Platform::MacOS, macos_path),
+        (Platform::MacCatalyst, mac_catalyst_path), (Platform::TvOS, tvos_path), (Platform::WatchOS, watchos_path),
+        (Platform::WatchOSSimulator, watchos_simulator_path), (Platform::CarPlayOS, carplay_path), (Platform::CarPlayOSSimulator, carplay_simulator_path)
+    ];
+    platforms.iter().for_each(|(platform, path)| {
+        if let Some(path) = path {
+            config.paths.insert(platform.clone(), path);
```
chinedufn, line 40: Don't need this `clone()` if we implement `Copy`

crates/swift-bridge-cli/src/clap_exec.rs
```diff
@@ -0,0 +1,47 @@
...
+    let out_dir = matches.value_of("out-dir").map(|p| Path::new(p)).unwrap(); // required
+    let name = matches.value_of("name").unwrap_or(out_dir.file_name().expect("Couldn't derive Swift Package name from output directory").to_str().unwrap());
+    
+    let mut config = CreatePackageConfig {
+        bridge_dir: &Path::new(bridges_dir),
+        paths: HashMap::new(),
+        out_dir: &out_dir,
+        package_name: name
+    };
+    let platforms = vec![
```
chinedufn, line 33:
```rust
let platforms = ApplePlatform::ALL;
for platform in platforms {
    if let Some(path) = matches.value_of(platform.as_dir_name()) {
        config.paths.insert(platform, path);
    }
}

// Where we define `ApplePlatform` ...
impl ApplePlatform {
    pub fn as_dir_name(&self) -> &'static str { ... }

    const ALL: &'static [Self] = &[ApplePlatform::Foo, ApplePlatform::Bar, ...];
}
```

chinedufn (changes requested):
Looks great. Just some minor feedback.

Can you rebase to a single commit with a good title and commit message?

i.e. `Add create-package command to CLI`

chinedufn:
Looks like the integration test suite got stuck https://github.com/chinedufn/swift-bridge/runs/5514191148?check_suite_focus=true? Any idea what happened here?

Jomy10:
> Looks like the integration test suite got stuck

I saw that too. I'm not sure what happened either since I did not add anything to the other packages except for a Debug derive. Maybe let's see if it happens again.

I'll rebase later.

crates/swift-bridge-cli/src/clap_exec.rs
Jomy10: Great idea!

Jomy10:
Alright, that should be everythin.

crates/swift-bridge-build/src/package.rs
```diff
@@ -9,23 +9,23 @@ use std::process::Command;
...
+    pub paths: HashMap<ApplePlatform, &'a P>,
     /// The directory where the package will be saved
-    pub out_dir: &'a dyn AsRef<Path>,
+    pub out_dir: &'a P,
     /// The name for the Swift package
     pub package_name: &'a str,
 }
 
-impl<'a> CreatePackageConfig<'a> {
+impl<'a, P: AsRef<Path> + ?Sized> CreatePackageConfig<'a, P> {
```
chinedufn, line 23: We don't want this generic bounds because it constraints every field to use the same monomorphized `P`

crates/swift-bridge-cli/src/clap_exec.rs
```diff
@@ -0,0 +1,41 @@
...
+        Some("create-package") => handle_create_package(matches.subcommand_matches("create-package").unwrap()),
+        _ => unreachable!("No subcommand or unknown subcommand given") // Shouldn't happen
+    }
+}
+
+/// Executes the `create-package` command
+fn handle_create_package(matches: &ArgMatches) {
+    let bridges_dir = matches.value_of("bridges-dir").unwrap(); // required
+    let out_dir = matches.value_of("out-dir").map(|p| Path::new(p)).unwrap(); // required
+    let name = matches.value_of("name").unwrap_or(out_dir.file_name().expect("Couldn't derive Swift Package name from output directory").to_str().unwrap());
```
chinedufn, line 18: Using the name of the directory as the package name feels a bit unexpected.

Can we start by just making name a required field and if in the future it ends up feeling redundant we can consider a shortcut like this?

crates/swift-bridge-cli/src/clap_exec.rs
```diff
@@ -0,0 +1,41 @@
...
+pub fn handle_matches(matches: ArgMatches) {
+    match matches.subcommand_name() {
+        Some("create-package") => handle_create_package(matches.subcommand_matches("create-package").unwrap()),
+        _ => unreachable!("No subcommand or unknown subcommand given") // Shouldn't happen
+    }
+}
+
+/// Executes the `create-package` command
+fn handle_create_package(matches: &ArgMatches) {
+    let bridges_dir = matches.value_of("bridges-dir").unwrap(); // required
```
chinedufn, line 16: If know that if we use the `#[derive(Debug, Parser)]` clap macro we should automatically get a nice help message if someone forgot a required field.

Is there an easy way to do that without using the macro?

If not, should we just use the macro? What do you think?

crates/swift-bridge-cli/src/clap_exec.rs
```diff
@@ -0,0 +1,41 @@
...
+        paths: HashMap::new(),
+        out_dir: &out_dir,
+        package_name: name
+    };
+    
+    let platforms = ApplePlatform::ALL;
+    let mut paths: HashMap<ApplePlatform, Box<&Path>> = HashMap::new();
+    for platform in platforms {
+        let path = matches.value_of(platform.dir_name());
+        let path = path.map(|p| Path::new(p));
```
chinedufn, line 31: We should be able to write this without all of this mapping and boxing, but no big deal for now.

chinedufn (changes requested):
Minor notes!

crates/swift-bridge-cli/src/clap_exec.rs
Jomy10: Using the Builder api has the same result, so this unwrap statement should never fail as Clap will always show a helpful eerror message.

Jomy10:
Ok, all of your notes should be addressed. 

I've changed the insertion of the platform paths to the best solution I could come up with.

chinedufn:
Great work!

PR #42, opened 2022-03-14
Explain why we prepend Rust imports

chinedufn:


chinedufn:
```
$ tree MySwiftPackage/
MySwiftPackage/
├── Package.swift
├── Sources
│   └── MySwiftPackage
│       ├── SwiftBridgeCore.swift
│       ├── __ReexportRustNativeLibrary.swift
│       └── my-rust-lib.swift
└── framework.xcframework
    ├── Info.plist
    ├── ios-arm64
    │   ├── Headers
    │   │   ├── SwiftBridgeCore.h
    │   │   ├── module.modulemap
    │   │   └── my-rust-lib.h
    │   └── libmy_rust_lib.a
    ├── ios-x86_64-simulator
    │   ├── Headers
    │   │   ├── SwiftBridgeCore.h
    │   │   ├── module.modulemap
    │   │   └── my-rust-lib.h
    │   └── libmy_rust_lib.a
    └── macos-x86_64
        ├── Headers
        │   ├── SwiftBridgeCore.h
        │   ├── module.modulemap
        │   └── my-rust-lib.h
        └── libmy_rust_lib.a
```

chinedufn:
@Jomy10 mind reviewing this ?

Jomy10:
Seems like a clean solution!

The only thing that is different from the current implementation is that methods defined in the xcframework can be called from outside the package.

So for example if I have a function that is exported:
```rust
fn print_hello() {
    printen!("Hello world")
}
```

Then in an executable package depending on our generated package, we would call it like so:
```swift
import MyPackage

print_hello() // Hello world
```

With this change, the following could also be done:
```swift
import MyPackage

print_hello() // Hello world

__swift_bridge__$print_hello() // Hello world
```

This is not a big deal since the method is preceded by `__`, but still something I thought I'd share.

chinedufn:
Ahhhh nice catch. Let's stick to the current approach then and not introduce this new approach.

I'll leave a comment in the code explaining why we prepend the imports.

chinedufn:
@Jomy10 mind taking another look?

Jomy10:
Looks good to me!

Issue #55, opened 2022-04-21
Xcode 13.3 requires artifact name matches target name for RustXcframework


simlay:
When running the `test-integration.sh` I see this at the end:
```
     Running `/Users/simlay/projects/infinyon/swift-bridge/target/debug/integration-test-create-swift-package`
objc[48421]: Class AppleTypeCRetimerRestoreInfoHelper is implemented in both /usr/lib/libauthinstall.dylib (0x1ee779eb0) and /Library/Apple/System/Library/PrivateFrameworks/MobileDevice.framework/Versions/A/MobileDevice (0x10816c4f8). One of the two will be used. Which one is undefined.
objc[48421]: Class AppleTypeCRetimerFirmwareAggregateRequestCreator is implemented in both /usr/lib/libauthinstall.dylib (0x1ee779f00) and /Library/Apple/System/Library/PrivateFrameworks/MobileDevice.framework/Versions/A/MobileDevice (0x10816c548). One of the two will be used. Which one is undefined.
objc[48421]: Class AppleTypeCRetimerFirmwareRequestCreator is implemented in both /usr/lib/libauthinstall.dylib (0x1ee779f50) and /Library/Apple/System/Library/PrivateFrameworks/MobileDevice.framework/Versions/A/MobileDevice (0x10816c598). One of the two will be used. Which one is undefined.
objc[48421]: Class ATCRTRestoreInfoFTABFile is implemented in both /usr/lib/libauthinstall.dylib (0x1ee779fa0) and /Library/Apple/System/Library/PrivateFrameworks/MobileDevice.framework/Versions/A/MobileDevice (0x10816c5e8). One of the two will be used. Which one is undefined.
objc[48421]: Class AppleTypeCRetimerFirmwareCopier is implemented in both /usr/lib/libauthinstall.dylib (0x1ee779ff0) and /Library/Apple/System/Library/PrivateFrameworks/MobileDevice.framework/Versions/A/MobileDevice (0x10816c638). One of the two will be used. Which one is undefined.
objc[48421]: Class ATCRTRestoreInfoFTABSubfile is implemented in both /usr/lib/libauthinstall.dylib (0x1ee77a040) and /Library/Apple/System/Library/PrivateFrameworks/MobileDevice.framework/Versions/A/MobileDevice (0x10816c688). One of the two will be used. Which one is undefined.
[...]

PR #56, opened 2022-04-21
Added arm64 to build in ci with lipo

### LINKED ISSUE #53: Support both ARM64 and x86-64 MacOS builds at once
If I supply `--macos target/aarch64-apple-darwin/debug/libmy.a` or `--macos target/x86_64-apple-darwin/debug/libmy.a` they work.

I'd like a way to supply both to build an XCFramework that'll work on the still supported and older Intel Macs.

simlay:
I think this can close #53?


chinedufn:
Thanks!

Would you mind changing the Swift Packages chapter use a universal build so that future people don't run into this? https://github.com/chinedufn/swift-bridge/blob/6b3cb07814140099191130737f29bffe3e8d7208/book/src/building/swift-packages/README.md

Then we should be good to go here.

SwiftRustIntegrationTestRunner/integration-test-create-swift-package/src/main.rs
```diff
@@ -8,7 +8,7 @@ fn main() {
     create_package(CreatePackageConfig {
         bridge_dir: PathBuf::from("swift-package-rust-library-fixture/generated"),
         paths: HashMap::from([
-            (Platform::MacOS, PathBuf::from("swift-package-rust-library-fixture/target/x86_64-apple-darwin/debug/libtest_swift_packages.a") as _),
+            (Platform::MacOS, PathBuf::from("swift-package-rust-library-fixture/target/universal/libtest_swift_packages.a") as _),
```
chinedufn, line 11: Hmm..

Wouldn't this change just be testing that universal builds work. 

Which would mean that we're adding an extra build in CI and adding that `lipo` call in order to just test that `lipo` actually works?

If so, maybe we just tweak the book to demonstrate a universal build and call it a day?

simlay: I actually did this because I wanted `./test-integration.sh` to run on my m1 mac. I messed about having the `integration-test-create-swift-package` subcrate use the `TARGET` triple that it's compiled o and then have the `build.sh` do `if [ $(uname -m) = "arm64" ]; then ...` but found this to be worse.

I concluded that just making the CI library just be universal solved this problem. I'm fine with either option. 

When I did #55, I had to change `x86_64` to `aarch64` in two places (and then not commit them) to verify the changes were correct before submitting the PR.

chinedufn: Ah got it thanks for explaining.

Looks like a clean debug build of the fixture on my machine takes about 9 seconds (mainly syn).

If we guesstimate that on the weaker CI hardware it takes like ... say 20 ... this change would add roughly 20 seconds to CI (back of napkin math).

So.. nothing that we need to worry about for now.

So this universal approach sounds good to me. 

If you don't mind tweaking the Swift Package section of the book we can land this CI tweak without changes.

simlay:
> Would you mind changing the Swift Packages chapter use a universal build so that future people don't run into this? https://github.com/chinedufn/swift-bridge/blob/6b3cb07814140099191130737f29bffe3e8d7208/book/src/building/swift-packages/README.md

Done! I made it also added the steps a universal iOS simulator build - `x86_64-apple-ios` and `aarch64-apple-ios-sim` as well.

chinedufn:
Awesome, thanks!

PR #70, opened 2022-04-26
[WIP] Add Result<T,U>

simlay:


simlay:
@chinedufn I've been playing with this off and on for a few days and think this PR will be huge if it's all in one. Do you have any suggestions on breaking it up into smaller pieces?

crates/swift-bridge-macro/src/lib.rs
```diff
@@ -30,5 +30,6 @@ mod ui_tests {
     fn ui() {
         let t = trybuild::TestCases::new();
         t.compile_fail("tests/ui/*.rs");
+        t.pass("tests/ui-pass/*.rs");
```
chinedufn, line 33: `crates/swift-integration-tests` can be used to make sure that things compile

Then (when applicable) we call the things from the `SwiftIntegrationTestRunner`

https://github.com/chinedufn/swift-bridge/tree/d05c6de93ac1c875adf25e2ed793d9dd86cb1f52/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests

chinedufn (commented):
Woah – glad you're working on this!

> Do you have any suggestions on breaking it up into smaller pieces?

I'd say a good start would be to add a single codegen test (in a `mod result_codegen_tests`) and then just implement just enough to make it pass.

Along with a `ResultTests.swift` that calls a function exposed in `swift-integration-tests/src/result.rs` so that we can ensure that whatever we tested in our codegen test actually works.

There are lots of codegen tests to use as examples:

https://github.com/chinedufn/swift-bridge/blob/607e30579ac4de1fab5ff66838253cf3d07ddfbb/crates/swift-bridge-ir/src/codegen/codegen_tests.rs#L31-L43

As well as Swift+Rust end to end tests to use as examples:

https://github.com/chinedufn/swift-bridge/tree/607e30579ac4de1fab5ff66838253cf3d07ddfbb/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests

---

Maybe the first PR just tests `fn some_function() -> Result<u8, u16>` which becomes `func some_function() throws -> u8` on the Swift side?

chinedufn:
Yeah definitely feel free to make very liberal use of `todo!()`.

Totally fine with only landing the tiniest amount of support for a first PR.

Let me know if I can help with anything!

chinedufn:
Some thoughts:

We need to generate an C header representation for a `Result<T,E>`. As well as a Rust `#[repr(C)]` representation.

Much like we do for transparent structs.

https://github.com/chinedufn/swift-bridge/blob/d05c6de93ac1c875adf25e2ed793d9dd86cb1f52/crates/swift-bridge-ir/src/codegen/codegen_tests/shared_struct_codegen_tests.rs#L113-L123

Because there is no `Result` on the Swift side and instead it's just `fn some_function() throws -> E` .. we don't need to worry about generating any new types on the Swift side.

Which means.. it doesn't really matter if we have lots of C Result representations for the same `T` and `E` combinations, since whoever is calling the function on the Swift side isn't actually engaging with the `Result` type.

---

Say we have this bridge module

```rust
#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        fn some_function() -> Result<u32, u8>;
    }
}
```

We would emit the following inside of the C header:

```c
typedef struct __swift_bridge__$some_function$Result$u32$u8 { bool is_ok; uint8_t ok; uint8_t err; } __swift_bridge__$Result$u32$u8__;
__swift_bridge__$some_function$Result$u32$u8 __swift_bridge__$some_function(void);
```

And then something like the following would be emitted inside of the generated Rust tokens:

```rust
#[repr(C)]
#[allow(non_camel_case_types)]
struct some_function_Result_u32_u8 {
    is_ok: bool,
    ok: u32,
    err: u8,
}
```

And the generated Swift would look something like:

```swift
func some_function() throws u8 {
    {let val = __swift_bridge__$some_function(); if val.is_ok { return val.ok } else { return val.err }}()
}
```

---

By the way.. the parser knows about everything in the bridge module.. So really we don't need to "namespace" the Result representation to the function .. but rather to the module ..

For example

```rust
#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        fn some_function() -> Result<u32, u8>;
        fn another_function() -> Result<u32, u8>;
    }
}
```

We don't need two different Result representations here.. We just need one that both share.. Since the `swift_bridge::bridge` proc macro can see both of them we can make sure that the codegen makes the final generated functions use the same `#[repr(C)]` result type and the same C header type.

To do this we would need the generated name for the Result type in the C header to be based on the crate name, path to the file that the module is in and the module name.

Totally fine to save this minor optimization for the future.

---

So... in terms of a first PR...

just generating the C header types for results that we see being returned from functions should be a great start.

chinedufn:
Yeah I'd say looking at the shared struct codegen and integration tests, as well as the Option codegen and integration tests should be helpful here.

chinedufn:
#111

Issue #100, opened 2022-06-03
Generate public initializer for shared structs


samnm:
When using swift-bridge in a Swift package it is not possible to initialize a shared struct from Swift due to the default initializer being internal.

I believe the only solution for this is to generate an initializer so that we can explicitly mark it as public.

chinedufn:
Ah nice catch.

# Steps to Fix

- Expose a shared struct in our swift package tests https://github.com/chinedufn/swift-bridge/blob/master/SwiftRustIntegrationTestRunner/swift-package-rust-library-fixture/src/lib.rs#L1-L6

- Add a test case that verifies that we can instantiate a shared struct from a Swift Package https://github.com/chinedufn/swift-bridge/blob/91b480250a59fee392b6f74c75682e4ee871bd99/SwiftRustIntegrationTestRunner/swift-package-test-package/Tests/swift-package-test-packageTests/swift_package_test_packageTests.swift#L6-L8

- Run tests using `./test-integration.sh`

- Tests will fail

- Add `public` to `crates/swift-bridge-ir` to make tests pass

samnm:
Wow that is incredibly helpful thank you. I will give it a shot today!

PR #101, opened 2022-06-04
Generate public initializers for shared structs

samnm:
Fixes https://github.com/chinedufn/swift-bridge/issues/100

Swift's automatic struct initializers default to internal, so I had to generate initializers in order to mark them as public.

In order to simplify some of the generation code I added a `StructField` trait which `NamedStructField` and `UnnamedStructField` both implement. I am not 100% sure if this aligns with your style preferences so feel free to suggest a different arrangement of all of that.

### DIFF
--- a/crates/swift-bridge-ir/src/codegen/codegen_tests/option_codegen_tests.rs
+++ b/crates/swift-bridge-ir/src/codegen/codegen_tests/option_codegen_tests.rs
@@ -846,10 +846,11 @@ mod shared_struct_with_option_field_ffi_repr {
         ExpectedSwiftCode::ContainsAfterTrim(
             r#"
 public struct SomeStruct {
+    public var field: Optional<UInt8>
+
     public init(field: Optional<UInt8>) {
         self.field = field
     }
-    public var field: Optional<UInt8>
 
     @inline(__always)
     func intoFfiRepr() -> __swift_bridge__$SomeStruct {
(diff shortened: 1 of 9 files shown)


crates/swift-bridge-ir/src/codegen/codegen_tests/option_codegen_tests.rs
```diff
@@ -846,7 +846,10 @@ mod shared_struct_with_option_field_ffi_repr {
         ExpectedSwiftCode::ContainsAfterTrim(
             r#"
 public struct SomeStruct {
-    var field: Optional<UInt8>
+    public init(field: Optional<UInt8>) {
```
chinedufn, line 849: Super minor.. Can we put the initializer after the fields and newline in between them?

My thinking here is that it's nice for anyone that's looking at a struct definition to be able to see the most important thing (what fields are available) first.

crates/swift-bridge-ir/src/codegen/generate_swift/shared_struct.rs
```diff
@@ -112,4 +104,75 @@ extension {option_ffi_name} {{
...
+        }
+
+        if !body.is_empty() {
+            body = format!("\n{}    ", body);
+        }
+
+        body
+    }
+
+    fn convert_fields<'a, T>(&self, struct_fields: impl IntoIterator<Item = &'a T>) -> String
```
chinedufn, line 154: Maybe `declare_fields`?

chinedufn (approved):
Looks great!

Also, thanks for doing some refactoring. 

The cleanup work is going to make it easier for the next person that touches this so I appreciate it.

I left a couple of minor comments. Feel free to accept or reject them (they aren't super important).

Let me know once you're ready and I'll merge this.

Thanks!

chinedufn:
Looks like CI is failing. Looks like you need to run `cargo fmt`.

Then you can run `cargo test --all` locally to see if all of the Rust tests are passing.

samnm:
All sorted.

Thanks for the quick review and for the fantastic library!

PR #105, opened 2022-07-27
Build examples in ci

chinedufn:


chinedufn:
See https://github.com/chinedufn/swift-bridge/pull/90

PR #111, opened 2022-09-20
Add support for Result<T, E>

chinedufn:
This commit adds suppoer for passing a `Result<T, E>` as an argument
from Rust -> Swift.

For example, the following is not possible:

```rust
use some_crate::SomeRustType;

#[swift_bridge::bridge]
mod ffi {
    extern "Swift" {
        fn swift_func_takes_callback(
            arg: Box<dyn FnOnce(Result<SomeRustType, String>)>,
        );
    }

    extern "Rust" {
        type SomeRustType;

        #[swift_bridge::init]
        fn new () -> SomeRustType;
    }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ In addition to allowing you to share your own custom structs, enums and classes 
 | *const T                                                        | UnsafePointer\<T>                                                |                                                                                    |
 | *mut T                                                          | UnsafeMutablePointer\<T>                                         |                                                                                    |
 | Option\<T>                                                      | Optional\<T>                                                     |                                                                                    |
-| Result\<T>                                                      |                                                                  | Not yet implemented                                                                |
+| Result\<T, E>                                                   | RustResult\<T, E>                                                |                                                                                    |
 | Have a Rust standard library type in mind?<br /> Open an issue! |                                                                  |                                                                                    |
 |                                                                 | Have a Swift standard library type in mind?<br /> Open an issue! |                                                                                    |
 <!-- ANCHOR_END: built-in-types-table -->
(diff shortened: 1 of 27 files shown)


simlay:
I think this supersedes #70 (sorry for taking too long), right?

chinedufn:
Yup!

This commit doesn't handle returning values though. Only arguments. So it's just one piece of the puzzle.

MOZGIII:
A quick search in the code shows some instances of `Result<T, E> is not yet supported` - is there a problem with implementing these, or were they just missed?

chinedufn:
We've been implementing support for things as we need them. 
So, there isn't a problem, they just haven't been implemented yet.
If you're interested in adding some I'd be very happy to help guide.

chinedufn:
Hey Antonito,

Nah no work has begun on that. I've briefly jotted down some ideas in the past.

https://github.com/chinedufn/swift-bridge/pull/70#pullrequestreview-953967571
https://github.com/chinedufn/swift-bridge/pull/70#issuecomment-1110253691

If anyone is interested in working on returning results I'd be happy to do a more thorough write up if the above wasn't enough.

Here's an example of how we implement returning from an async function (not a result though). https://github.com/chinedufn/swift-bridge/blob/dc1119e1ee333e1412f81f726987637b8f620a58/crates/swift-bridge-ir/src/codegen/codegen_tests/async_function_codegen_tests.rs#L182-L268

And here's the guide for contributing new signatures https://chinedufn.github.io/swift-bridge/contributing/adding-support-for-a-signature/index.html

Antonito:
I started digging into it, I'm not sure if my approach is correct @chinedufn – is there any way we could chat over IM (a Slack, Discord ?) to make the process smoother? :)
I just opened my Twitter DM (on my Github profile) if you want to reach out to exchange details

chinedufn:
I don't have any IM or Twitter, but I'd be down to hop in one temporarily to chat (I'd prefer a call, but IM works too).
Want to email me (email in GH profile) and we can coordinate a way to chat?

Thanks for diving in!!

Antonito:
Sounds good, I just emailed you – thank you very much!

chinedufn:
For posterity, @NiwakaDev did some AMAZING work in #157 to land support for `async fn foo() -> Result<OpaqueRust, OpaqueRust`.

There's still some more work to do to make things seamless such as #150  .. but if anyone in this thread was blocked by opaque Rust types inside of results it should be _possible_ after #157 .

PR #121, opened 2022-12-13
Fix minor swift-bridge-cli incantation to use the correct ios binary

bes:


chinedufn:
I think this depends on whether or not you're on `x86` or `arm`?

bes:
`x86_64-apple-ios` and `aarch64-apple-ios-sim` are lipo'd together to cover simulators, right? It doesn't make sense to run x86 as an ios target? Anyway when I run the example as-is (before my fix) it complains:

```shell
Both 'ios-x86_64-simulator' and 'ios-arm64_x86_64-simulator' represent two equivalent library definitions.
thread 'main' panicked at '', /Users/bes/.cargo/registry/src/github.com-1ecc6299db9ec823/swift-bridge-build-0.1.41/src/package.rs:239:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

chinedufn:
Oh right, x86 is only for simulators. Makes sense.

Thanks for fixing this!

Issue #122, opened 2022-12-14
Support `swift_name` attribute on extern "Swift" functions.


bes:
Hi! First - thank you for this project! I have successfully called Rust code from Swift!

Unfortunately, I am having some troubles, and I can't figure out how to fix them.

I want to call a Swift function from Rust, but the generated swift code has a compiler error in it.

This is the rust code (snippet)

```rust
#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        // Works, except the function is still called rust_init on the Swift side?
        #[swift_bridge(swift_name = "rustInit")]
        fn init(url: String);
    }

    extern "Swift" {
        // Generates a Swift compiler error
        #[swift_bridge(swift_name = "getAuthToken")]
        fn get_authentication_token() -> String;
    }
}
```

This snippet is generated

```swift
@_cdecl("__swift_bridge__$get_authentication_token")
func __swift_bridge__get_authentication_token () -> RustString {
    { let rustString = getAuthToken().intoRustString(); rustString.isOwned = false; return rustString.ptr }()
}
```

And then I added this function (in the Swift package, as a new file)

```swift
public func getAuthToken() -> String {
    return "token"
}
```

But the generated snippet has these compiler errors: 

* `Method cannot be marked @_cdecl because its result type cannot be represented in Objective-C`
* `Cannot convert return expression of type 'UnsafeMutableRawPointer' to return type 'RustString'`

An additional, minor, problem is that `#[swift_bridge(swift_name = "rustInit")]` doesn't seem to do anything.

What am I doing wrong? Thanks.

chinedufn:
As for the `swift_name`.

We parse it here https://github.com/chinedufn/swift-bridge/blob/3018fc2d051b34ece0f4709d78c7deb9bf4bb01c/crates/swift-bridge-ir/src/parse/parse_extern_mod/function_attributes.rs#L89-L94

But we don't actually use it.

To fix this, we would first add an integration test. We can use the `rust_name` one as inspiration.

https://github.com/chinedufn/swift-bridge/blob/78be911bbefc847a0ce4e7dfdeaf2deda68c408c/crates/swift-integration-tests/src/function_attributes.rs#L7

https://github.com/chinedufn/swift-bridge/blob/ff3852c6198a43ca3550348be04584f4914c9305/crates/swift-integration-tests/src/function_attributes/rust_name.rs#L1-L10

[...]

PR #123, opened 2022-12-14
Implement return String from Swift to Rust

bes:
Attempts to fix one part of https://github.com/chinedufn/swift-bridge/issues/122

### DIFF
--- a/crates/swift-bridge-ir/src/codegen/codegen_tests/string_codegen_tests.rs
+++ b/crates/swift-bridge-ir/src/codegen/codegen_tests/string_codegen_tests.rs
@@ -215,3 +215,50 @@ struct RustStr __swift_bridge__$some_function(void);
         .test();
     }
 }
+
+/// Test code generation for Swift function that returns an owned String argument.
+mod extern_swift_func_returns_string {
+    use super::*;
+
+    fn bridge_module_tokens() -> TokenStream {
+        quote! {
+            mod foo {
+                extern "Swift" {
+                    fn some_function () -> String;
+                }
+            }
+        }
+    }
+
+    fn expected_rust_tokens() -> ExpectedRustTokens {
+        ExpectedRustTokens::Contains(quote! {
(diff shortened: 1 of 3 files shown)


crates/swift-bridge-ir/src/codegen/codegen_tests/string_codegen_tests.rs
```diff
@@ -215,3 +215,50 @@ struct RustStr __swift_bridge__$some_function(void);
...
+                }
+            }
+        }
+    }
+
+    fn expected_rust_tokens() -> ExpectedRustTokens {
+        ExpectedRustTokens::Contains(quote! {
+            pub fn some_function () -> String {
+                unsafe {
+                    Box::from_raw(unsafe { __swift_bridge__some_function() }).0
```
bes, line 237: Is this really safe? You're assuming that Rust can free the memory safely, when the Swift runtime and the Rust runtime can technically have different allocators? http://jakegoulding.com/rust-ffi-omnibus/string_return/

chinedufn: Yes it should be safe because:

1. `__swift_bridge__$some_function` creates a `RustString`, which is a Swift class that holds a pointer to a new type wrapper around a `std::string::String` https://github.com/chinedufn/swift-bridge/blob/23d60feaa0f8e92f2c854d8665868aa5573ef505/src/std_bridge/string.rs#L22-L23

2. Pointer to the Rust `std_bridge::string::RustString` gets passed from Swift -> Rust.

3. Rust uses `Box::from_raw` to get the `RustString(std::string::String)`

4. Rust now has ownership of the `String`.

So, the `RustString` is only ever allocated and freed from Rust's allocator.

---

Let me know if this doesn't answer your question.

chinedufn (approved):
Great stuff. Thanks for this awesome contribution!!!

PR #126, opened 2022-12-15
Make enums support the `swift_name` attrib

chinedufn:
This commit makes enums support the `swift_name` attribute, which lets
you choose the name that shows up on the Swift side.

For example, the following is now possible:

```rust
#[swift_bridge::bridge]
mod ffi {
    #[swift_bridge(swift_name = "EnumRename")]
    enum EnumName {
        Variant1,
        Variant2,
    }
}
```


bes:
This should complete everything I asked about in #122 ?

chinedufn:
This doesn't add support for `swift_name` on Rust functions, so #122 is still open.

We essentially need to change this https://github.com/chinedufn/swift-bridge/blob/54b9c09337701dd7580010d1d5ef6527b855ee98/crates/swift-bridge-ir/src/codegen/generate_swift/generate_function_swift_calls_rust.rs#L62

To something along the lines of:

```rust
if let Some(swift_name) = &function.swift_name_override {
    format!("public func {}", swift_name.value())
} else {
    format!("public func {}", fn_name.as_str())
}
```

PR #131, opened 2023-01-05
Make RustVec methods public

onato:
We can't use these methods from outside of the framework unless they are public.


chinedufn:
Thanks!

PR #132, opened 2023-01-08
Fix README.md in swiftc-and-cargo

NiwakaDev:
When I tried to build a project following swiftc+Cargo page of the book, I got the following error.
```
generated/swift-and-rust/swift-and-rust.swift:6:5: error: cannot find 'print_hello_swift' in scope
print_hello_swift()
^~~~~~~~~~~~~~~~~
```
To resolve the above error, I implemented a ```print_hello_swift``` function.

In addition,  I added a procedure to change the permissions of build-swiftc-links-rust.sh.

### DIFF
--- a/book/src/building/swiftc-and-cargo/README.md
+++ b/book/src/building/swiftc-and-cargo/README.md
@@ -60,6 +60,10 @@ run()
 
 ```swift
 // lib.swift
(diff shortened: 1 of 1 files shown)


chinedufn:
Thanks!

Issue #135, opened 2023-01-12
Xcode 14: swift package setup doesn't work


akesson:
Thanks for creating swift-bridge. It looks like exactly what I need, but Xcode is giving me hell...

I followed the [instructions](https://chinedufn.github.io/swift-bridge/building/swift-packages/index.html) but when trying to run the project I get a build error `No such module 'PolyglotServer'`

When adding the package dependency it ends up show as a folder instead of a swift package. 
<img width="872" alt="Captura de pantalla 2023-01-12 a las 17 02 39" src="https://user-images.githubusercontent.com/1822029/212119365-e5cbfdf7-2f9e-41ca-8abf-61ff160f444a.png">

Using Xcode 14.2 and swift-bridge 0.1.45

akesson:
FYI: I'm on Mac M1.

The Xcode + Cargo instructions are really good. They worked, with one exception, it only works on a real phone. 

Seems like cargo-lipo doesn't build the simulator binaries. 
```
[INFO  cargo_lipo::meta] Will build universal library for ["ios-rust-analyzer"]
[INFO  cargo_lipo::lipo] Building "ios-rust-analyzer" for "aarch64-apple-ios"
   Compiling ios-rust-analyzer v0.1.0 (/Users/hakesson/Developer/evaluate/ios-rust-analyzer)
    Finished dev [unoptimized + debuginfo] target(s) in 4.01s
[INFO  cargo_lipo::lipo] Building "ios-rust-analyzer" for "x86_64-apple-ios"
   Compiling ios-rust-analyzer v0.1.0 (/Users/hakesson/Developer/evaluate/ios-rust-analyzer)
    Finished dev [unoptimized + debuginfo] target(s) in 2.22s
[INFO  cargo_lipo::lipo] Creating universal library for ios-rust-analyzer
```

When trying to run on sim I get:

> In /Users/hakesson/Developer/evaluate/ios-rust-analyzer/target/universal/debug/libios_rust_analyzer.a(ios_rust_analyzer-ec4a45b193521035.103my0zmix07evm9.rcgu.o), building for iOS Simulator, but linking in object file built for iOS, file '/Users/hakesson/Developer/evaluate/ios-rust-analyzer/target/universal/debug/libios_rust_analyzer.a' for architecture arm64

akesson:
Did a PR for the simulator/device building: https://github.com/chinedufn/swift-bridge/pull/136

chinedufn:
[...]

PR #136, opened 2023-01-12
Check which TARGET to build: simulator or device ones.

akesson:
Note: I only tested this on a mac M1

### DIFF
--- a/book/src/building/xcode-and-cargo/README.md
+++ b/book/src/building/xcode-and-cargo/README.md
@@ -85,18 +85,28 @@ if [[ -n "${DEVELOPER_SDK_DIR:-}" ]]; then
   export LIBRARY_PATH="${DEVELOPER_SDK_DIR}/MacOSX.sdk/usr/lib:${LIBRARY_PATH:-}"
 fi
 
+if [[ $LLVM_TARGET_TRIPLE_SUFFIX == "-simulator" ]]; then
+    if [[ $NATIVE_ARCH == "arm64" ]]; then
+        export TARGETS="aarch64-apple-ios-sim"
+    else
+        export TARGETS="x86_64-apple-ios-sim"
+    fi
+else
+    export TARGETS="aarch64-apple-ios,x86_64-apple-ios"
+fi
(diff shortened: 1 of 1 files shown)


chinedufn:
Hey, thanks!

What problem is trying to solve?

Right now we build a universal binary that should run on both native devices and the simulator. It seems like you're suggesting that that isn't working?

akesson:
Please see: https://github.com/chinedufn/swift-bridge/issues/135#issuecomment-1380747193

book/src/building/xcode-and-cargo/README.md
```diff
@@ -85,18 +85,24 @@ if [[ -n "${DEVELOPER_SDK_DIR:-}" ]]; then
   export LIBRARY_PATH="${DEVELOPER_SDK_DIR}/MacOSX.sdk/usr/lib:${LIBRARY_PATH:-}"
 fi
 
+if [[ $LLVM_TARGET_TRIPLE_SUFFIX == "-simulator" ]]; then
```
chinedufn, line 88: Just so that we don't pollute the environment vars of sub-processes:

```sh
TARGETS=""
if [ "$PLATFORM_NAME" = "iphonesimulator" ]; then
  TARGETS="..."
else
  TARGETS="..."
fi
```

chinedufn (approved):
Ok so if I'm understanding correctly

- `cargo lipo` defaults to `aarch64-apple-ios,x86_64-apple-ios`

- On M1 macs the simulator runs `aarch64-apple-ios-sim`

- So, we need to build for `aarch64-apple-ios-sim` when building for the simulator

So if CI passes (CI runs on intel mac) then we're good to go. Otherwise we'll have to look into a way for this to all work on intel and m1.

chinedufn:
Requested one minor change then LGTM.

Thanks!

book/src/building/xcode-and-cargo/README.md
akesson: Yes, you are right!

Issue #129, opened 2022-12-18
Document how to add Swift files to a generated Swift Package


bes:
I would like to read some documentation on what you think the best place for auxillary Swift files for the generated Swift Package is.

Might it be
* In the directory of the "target Swift Package" itself (could be another folder / git)?
* In some directory in the "source crate" using `swift-bridge`?
* Some other place
* Undefined

Is there some automatic facility for copying Swift files already in place?

chinedufn:
Is this related https://github.com/chinedufn/swift-bridge/pull/138 ?

bes:
I don't think so. But I might be wrong.

Since `swift-bridge` generates a Swift package, and since I then import the Swift package from my XCode project, I don't have access to the code inside the larger XCode project from the Swift package generated by `swift-bridge`.

To solve that, I use inversion-of-control - I declare a `public protocol Proxy` inside the Swift package, and then the main application can bind to the Swift package like so:

```swift
public protocol Proxy {
    // ...
}

var proxy Proxy?

public func registerProxy(_ newProxy: Proxy) {
    proxy = newProxy
}
```

So basically this requires custom code inside the Swift package generated by `swift-bridge`, but there is no documentation on how to maintain / manage that code.

Some options:

* Maybe the Swift proxy code should live inside the Rust project?
* Maybe the Swift proxy code should live inside the generated Swift project?
* Or maybe `swift-bridge` doesn't care, and you can do whatever you like.

Either way, the most preferred way should be documented, maybe?

chinedufn:
> I don't have access to the code inside the larger XCode project from the Swift package generated by swift-bridge

I see, thanks for the example code.

---

I'd like to better understand your use case so that we can figure out the best solution.

Can you explain what you're trying to do, exactly?
Why does your Swift package need to know about the project that is using it?

bes:
Sure —

Here's the actual `Proxy` protocol

```swift
public protocol Proxy {
    func getAuthToken() -> String
    func getApiUrl() -> String
    func log(line: String)
}
```

The main Swift application supplies the Rust library with (amongst other things) an authentication token (on demand with short lifetime), the URL to the API server (can be different depending on dev/prod/region) and a logging facility so that I can easily mix Swift and Rust logs.

bes:
To spell it out even more, the sequence of events would be:

Rust needs auth token -> Calls getAuthToken through swift-bridge -> ends up in Swift package -> Swift package uses Proxy protocol to get actual token from main app

[...]

PR #138, opened 2023-01-13
Add resources to package

akesson:
Added a `--resource` option to the cli which copies resources into the package and adds the appropriate Bundle extension.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,3 +40,4 @@ members = [
   "examples/codegen-visualizer",
   "examples/rust-binary-calls-swift-package",
 ]
+default-members = ["crates/swift-bridge-cli"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,4 +40,3 @@ members = [
   "examples/codegen-visualizer",
   "examples/rust-binary-calls-swift-package",
 ]
-default-members = ["crates/swift-bridge-cli"]
--- a/SwiftRustIntegrationTestRunner/integration-test-create-swift-package/src/main.rs
+++ b/SwiftRustIntegrationTestRunner/integration-test-create-swift-package/src/main.rs
@@ -15,5 +15,6 @@ fn main() {
         )]),
         out_dir: PathBuf::from("swift-package-rust-library-fixture/MySwiftPackage"),
         package_name: "MySwiftPackage".to_string(),
+        resources: vec![],
     });
 }
--- a/crates/swift-bridge-build/src/package.rs
+++ b/crates/swift-bridge-build/src/package.rs
@@ -329,17 +329,7 @@ extension Bundle {{
     fs::write(bridge_path, bridge_content)
         .expect("Couldn't copy project's bridging swift file to the package");
 
-    // Copy resources
-    let mut resource_entries: Vec<String> = vec![];
-    for (from, to) in &config.resources {
-        resource_entries.push(format!("				.copy(\"{}\")", &to.display()));
-        let to = sources_dir.join(to);
-
-        if let Some(parent) = to.parent() {
-            fs::create_dir_all(parent).expect("Couldn't create directory for resource");
-        }
-        fs::copy(from, to).expect("Couldn't copy resource");
-    }
+    let resource_entries = copy_resources(&config.resources, &sources_dir);
 
     // Generate Package.swift
     let package_name = &config.package_name;
@@ -383,4 +373,18 @@ fn first_lowercased(name: &str) -> String {
         None => String::new(),
         Some(f) => f.to_lowercase().collect::<String>() + c.as_str(),
     }
+}
+
+fn copy_resources(resources: &[(PathBuf, PathBuf)], sources_dir: &Path) -> Vec<String> {
+    let mut resource_entries: Vec<String> = vec![];
+    for (from, to) in resources {
+        resource_entries.push(format!("				.copy(\"{}\")", &to.display()));
+        let to = sources_dir.join(to);
+
+        if let Some(parent) = to.parent() {
+            fs::create_dir_all(parent).expect("Couldn't create directory for resource");
+        }
+        fs::copy(from, to).expect("Couldn't copy resource");
+    }
+    resource_entries
 }
--- a/crates/swift-bridge-build/src/package.rs
+++ b/crates/swift-bridge-build/src/package.rs
@@ -313,6 +313,18 @@ fn gen_package(output_dir: &Path, config: &CreatePackageConfig) {
     )
     .expect("Couldn't copy project's bridging swift file to the package");
 
+    // Copy resources
+    let mut resource_entries: Vec<String> = vec![];
+    for (from, to) in &config.resources {
+        resource_entries.push(format!("				.copy(\"{}\")", &to.display()));
+        let to = sources_dir.join(to);
+
+        if let Some(parent) = to.parent() {
+            fs::create_dir_all(parent).expect("Couldn't create directory for resource");
+        }
+        fs::copy(from, to).expect("Couldn't copy resource");
+    }
+
     // Generate Package.swift
     let package_name = &config.package_name;
     let package_swift = format!(
@@ -333,10 +345,15 @@ let package = Package(
 		),
 		.target(
 			name: "{package_name}",
-			dependencies: ["RustXcframework"])
+			dependencies: ["RustXcframework"],
+			resources: [
+{}
+			]
+		)
 	]
 )
-	"#
+	"#,
+        resource_entries.join(",\n")
     );
 
     fs::write(output_dir.join("Package.swift"), package_swift)
--- a/crates/swift-bridge-build/src/package.rs
+++ b/crates/swift-bridge-build/src/package.rs
@@ -381,10 +381,16 @@ fn copy_resources(resources: &[(PathBuf, PathBuf)], sources_dir: &Path) -> Vec<S
         resource_entries.push(format!("				.copy(\"{}\")", &to.display()));
         let to = sources_dir.join(to);
 
-        if let Some(parent) = to.parent() {
-            fs::create_dir_all(parent).expect("Couldn't create directory for resource");
+        if from.is_dir() {
+            copy_dir::copy_dir(from, &to).expect(&format!(
+                "Could not copy resource directory {from:?} to {to:?}"
+            ));
+        } else {
+            if let Some(parent) = to.parent() {
+                fs::create_dir_all(parent).expect("Couldn't create directory for resource");
+            }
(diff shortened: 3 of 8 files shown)


chinedufn:
Thanks a lot for working on this!

Can you explain what led you to wanting this feature and how/when you expect to use it?

This way we can evaluate the design and feel confident that we're taking the best approach.

---

I don't have much Swift experience and I'm not familiar with bundle extensions. 
It this a good summary for me to ready -> https://developer.apple.com/documentation/foundation/bundle ?

akesson:
> Can you explain what led you to wanting this feature and how/when you expect to use it?

I'm creating a product which uses [Leptos](https://crates.io/crates/leptos) for building a server binary & client-side wasm and I want to package that into a Swift package in order to use it in a WebView-based iOS/macOS app. The server binary needs the site data (css, js, wasm etc) accessible, which is where this feature comes in handy.

> I don't have much Swift experience and I'm not familiar with bundle extensions.

You make me curious. Why are you creating this (really good) swift-bridge? 

See [Bundling resources with a Swift package](https://developer.apple.com/documentation/xcode/bundling-resources-with-a-swift-package) and [SPM Resource](https://github.com/apple/swift-package-manager/blob/main/Documentation/PackageDescription.md#resource)

crates/swift-bridge-build/src/package.rs
```diff
@@ -18,6 +18,8 @@ pub struct CreatePackageConfig {
     pub out_dir: PathBuf,
     /// The name for the Swift package
     pub package_name: String,
+    /// Additional resources to copy into the package, first PathBuf is source and second is destination
+    pub resources: Vec<(PathBuf, PathBuf)>,
```
chinedufn, line 22: Can we make the semantics of the destination more clear?
Right now it's hard to tell what the destination is.

 Maybe `pub resources: Vec<BundleResource>`

```rust
/// Describes a bundle resource to copy from some source path to some destination.
///
/// The source is a path to an existing file or directory that you wish to bundle.
///
/// The destination is a relative path that the resource will get copied to within the
/// generated Swift package's `Sources/YourLibraryName` directory.
///
/// # Examples
///
/// ```
/// # use crate::CopyBundleResourceDesc;
///
/// // If you are creating a Swift Package called "MyLibrary", this descriptor will
/// // lead to the package generator copying the `/path/to/images` directory to the
/// // the Swift package's `Sources/MyLibrary/assets/icons` directory.
/// let resource = CopyBundleResourceDesc::new("/path/to/images", "assets/icons");
/// ```
pub struct CopyBundleResourceDesc {
    source: PathBuf,
    destination: PathBuf,
}

impl CopyBundleResourceDesc {
    /// See [`CopyBundleResourceDesc`] for documentation.
    ///
    /// # Panics
    /// Panics if the destination is an absolute path.
    pub fn new(source: impl Into<PathBuf>, destination: impl Into<PathBuf>) -> Self {
        let destination = destination.into();
        assert!(destination.is_relative());
        Self {
            source: source.into(),
            destination,
        }
    }
}
```

Just a quick idea, open to other approaches.

SwiftRustIntegrationTestRunner/integration-test-create-swift-package/src/main.rs
```diff
@@ -15,5 +15,6 @@ fn main() {
         )]),
         out_dir: PathBuf::from("swift-package-rust-library-fixture/MySwiftPackage"),
         package_name: "MySwiftPackage".to_string(),
+        resources: vec![],
```
chinedufn, line 18: Can we add some resources?

- A file `resource-test.txt`
- A directory with 2 files `directory-with-files/{a,b}.txt
- A directory that contains a directory that contains a file `nested/directory/test.txt`

Then right after this `create_package` call we can call a `fn assert_resources_bundled`

Which checks the "swift-package-rust-library-fixture/MySwiftPackage" Sources dir to make sure that the resources are in there.

---

Also, if there's a way to check that the resources are accessible on the Swift side, that would be great. Could check in here https://github.com/chinedufn/swift-bridge/blob/a207c24d328b938f01642befe9c2be03ae3a7c5b/SwiftRustIntegrationTestRunner/swift-package-test-package/Tests/swift-package-test-packageTests/swift_package_test_packageTests.swift#L5-L17

Cargo.toml
```diff
@@ -40,3 +40,4 @@ members = [
   "examples/codegen-visualizer",
   "examples/rust-binary-calls-swift-package",
 ]
+default-members = ["crates/swift-bridge-cli"]
```
chinedufn, line 43: Can we remove this?

crates/swift-bridge-build/src/package.rs
```diff
@@ -298,15 +302,33 @@ fn gen_package(output_dir: &Path, config: &CreatePackageConfig) {
...
-                .expect("Couldn't read project's bridging swift file")
-        ),
-    )
-    .expect("Couldn't copy project's bridging swift file to the package");
+            r#"import RustXcframework
+import Foundation
+
+{bridge_swift}
+
+extension Bundle {{
```
chinedufn, line 318: What does this do?

crates/swift-bridge-cli/src/clap_app.rs
```diff
@@ -100,4 +100,16 @@ fn create_package_command() -> Command<'static> {
                 .required(true)
                 .help("The name for the Swift Package"),
         )
+        .arg(
+            Arg::new("resource")
+                .long("resource")
+                .takes_value(true)
+                .value_name("SRC_PATH:DEST_PATH")
+                .multiple_values(true)
+                .help(
```
chinedufn, line 109: Can we update the book example to include bundling a resource?

https://github.com/chinedufn/swift-bridge/blob/2783798ef267135aeafdbb8981c7ff93efb0dfb8/book/src/building/swift-packages/README.md#L132-L171

crates/swift-bridge-cli/src/clap_exec.rs
```diff
@@ -34,3 +36,30 @@ fn handle_create_package(matches: &ArgMatches) {
...
+fn parse_resource_pairs(matches: &ArgMatches) -> Vec<(PathBuf, PathBuf)> {
+    let pairs: Vec<String> = matches
+        .get_many("resource")
+        .map(|v| v.cloned().collect())
+        .unwrap_or_default();
+
+    pairs
+        .into_iter()
+        .map(|pair| {
+            let mut split = pair.split(':');
```
chinedufn, line 49: Can we leave a TODO comment that this breaks for paths that contain `:`, i.e. "/path/to/my\:file.txt".

If we're going to eventually switch to https://github.com/clap-rs/clap/issues/1682 then not worrying about this bug is fine since it won't be a problem when we switch and it's likely that way less than 1% of people would ever run into this.

crates/swift-bridge-build/src/package.rs
```diff
@@ -328,12 +350,46 @@ let package = Package(
 		),
 		.target(
 			name: "{package_name}",
-			dependencies: ["RustXcframework"])
+			dependencies: ["RustXcframework"],
+			resources: [
```
chinedufn, line 354: Can we move the package string generation out into its own function https://github.com/chinedufn/swift-bridge/blob/03de7ea4fb646de1b71358a717488ddac5d67e97/crates/swift-bridge-build/src/package.rs#L313-L335

Then add a test to the bottom of this file that generates the package description for the `CreatePackageConfig` that has resources and asserts on the contents of the generated string?

My main concern here is that right now the writing to disk / process spawning stuff is tightly coupled to things that it doesn't need to be (I know this was here before you, so not saying that you did that!).

So I'm asking to add a test for this resources thing so that we at least begin take a small step in the right direction and make it easier to make our Swift package creation more powerful over time.

chinedufn (commented):
This is awesome, thanks so much for contributing this.

I think your work here will close https://github.com/chinedufn/swift-bridge/issues/129 .

I left some questions and a few comments about adding tests. After those are addressed this LGTM and I can merge and cut a new release.

Thanks for your great work!

chinedufn:
https://github.com/chinedufn/swift-bridge/pull/138#issuecomment-1382825188

Ok thanks for explaining and thanks for the links. Read them. They were helpful.

> You make me curious. Why are you creating this (really good) swift-bridge?

Thanks for the kindness, this made me smile. 
I maintain an iOS app that uses SwiftUI for views and Rust for all of the logic. I quickly got tired of passing around pointers, and eventually came across [cxx](https://github.com/dtolnay/cxx) and got inspired to try the "bridge module" approach.

crates/swift-bridge-cli/src/clap_app.rs
```diff
@@ -100,4 +100,16 @@ fn create_package_command() -> Command<'static> {
                 .required(true)
                 .help("The name for the Swift Package"),
         )
+        .arg(
+            Arg::new("resource")
```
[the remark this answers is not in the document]
akesson: TBH, it's not a great solution to use `:`, but I didn't really find any easy and good solution so I went for the easy one :)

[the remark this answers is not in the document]
akesson: Added the TODO

crates/swift-bridge-build/src/package.rs
akesson: Ok, refactoring into CopyBundleResourceDesc done.

crates/swift-bridge-build/src/package.rs
akesson: I did make an effort to follow your (current) style of programming, imitating how things are done. I think it is better if you, as the owner of this project, set the new precedence instead of me as I will probably not do it as you like it.

crates/swift-bridge-build/src/package.rs
akesson: Added comment in code

Cargo.toml
akesson: Done

crates/swift-bridge-cli/src/clap_app.rs
akesson: I added it to the end as it is not mandatory to add resources.

akesson:
I'm afraid that I won't have more time to spend on this. Feel free to reject the PR if it is not up to your standards.

crates/swift-bridge-build/src/package.rs
chinedufn: Actually all of the package stuff was contributed by a contributor.

However, makes sense to me!

chinedufn:
Ok thanks for the heads up.

It's just missing tests. I can take get them added when I have time. Thanks for your work!

PR #143, opened 2023-01-25
Remove obsolete println! in build process

### LINKED ISSUE #142: Add an option to run Swift build functions silently (without any output to stdout)
For usage in dedicated build / bundling tools, it would be handy to configure or use the swift_bridge_build functions silently, without producing output to stdout (being aware of the drawbacks, this includes cargo-commands (cargo-rerun-if, etc.)).

A few possible solutions:
- crate feature flag (e.g. silent) that is disabled by default (+ does not bother users who do not need this, - makes the setting global, so no fine-grained control possible)
- introduce additional parameter to functions (+ fine-grained control, - bothers everybody, even if this is not needed / not advisable for people who use swift_bridge_build in a build.rs file)
- introduce a silent function for every function (+ fine-grained control, - refactoring / maintainance and probably migration overhead, - invoking the wrong version of a function accidentally is possible)
- provide no solution for this. If using swift_bridge_build in separate tools instead of invoking it in build.rs is strongly discouraged, this is also an option to consider.

Context: I am building [cargo-swift](https://github.com/antoniusnaumann/cargo-swift) that is intended to be an opinionated, user-friendly convenience layer above swift_bridge, mainly aiding in setting up boilerplate and building a Swift package from your codebase without requiring any extra build configuration (steps that would be required in build.rs (e.g. generating bridging header) as well as scripts such as lipo. It therefore pursues other goals than the swift-bridge-cli which enables more fine-grained control but requires setting up a custom project build config and external tools (e.g. lipo) to be invoked manually)

I am willing to work on this feature after discussing how it should be integrated. I am in favor of providing this as a crate feature that is disabled by default.

antoniusnaumann:
This PR removes "cargo:rerun-if-changed" statements that were no longer used.
It does not include a solution to silence xcodebuild output, since that particular output of xcodebuild -create-xcframework does not respect the global -quiet option.

Closes #142

### DIFF
--- a/crates/swift-bridge-build/src/generate_core.rs
+++ b/crates/swift-bridge-build/src/generate_core.rs
@@ -2,7 +2,7 @@ use crate::generate_core::boxed_fn_support::{
     C_CALLBACK_SUPPORT_NO_ARGS_NO_RETURN, SWIFT_CALLBACK_SUPPORT_NO_ARGS_NO_RETURN,
 };
 use crate::generate_core::result_support::{C_RESULT_SUPPORT, SWIFT_RUST_RESULT};
-use std::path::{Path, PathBuf};
+use std::path::Path;
 
 const RUST_STRING_SWIFT: &'static str = include_str!("./generate_core/rust_string.swift");
 const RUST_STRING_C: &'static str = include_str!("./generate_core/rust_string.c.h");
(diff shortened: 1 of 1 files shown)


chinedufn:
Thanks!

PR #144, opened 2023-01-25
Capture output from xcodebuild instead of redirecting it to parent's stdout

antoniusnaumann:
I verified that this removes the output in question.


chinedufn:
Thanks!

Issue #11, opened 2022-01-17
Introduce a compile time error when using missing argument in `args_into = (...)` attribute


chinedufn:
```rust
#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        // Right now this compiles, but we want to instead get
        // a compile time error saying that `arg_typo` is not an
        // argument.
        #[swift_bridge(args_into = (arg_typo))]
        fn some_function(arg: u8);
    }
}
```

chinedufn:
# Guide

### Add a UI test

We should add a UI test, perhaps called `args-into-argument-not-found.rs` and `args-into-argument-not-found.stderr`

Here's an example of a different UI test:

https://github.com/chinedufn/swift-bridge/blob/6a10cc2d28f07c36c4ade564cc49f4eaeb787eed/crates/swift-bridge-macro/tests/ui/invalid-module-item.rs#L1-L10

https://github.com/chinedufn/swift-bridge/blob/6a10cc2d28f07c36c4ade564cc49f4eaeb787eed/crates/swift-bridge-macro/tests/ui/invalid-module-item.stderr#L1-L11

More UI tests can be found here -> https://github.com/chinedufn/swift-bridge/tree/6a10cc2d28f07c36c4ade564cc49f4eaeb787eed/crates/swift-bridge-macro/tests/ui

Perhaps our args into UI test has a function with one missing arg, and another function with two missing args.

### ParseError

We can add a new `ParseError` variant for this scenario. 
Perhaps something like `ParseError::ArgsIntoArgNotFound { func: ForeignItemFunc, missing_arg: Ident }`

https://github.com/chinedufn/swift-bridge/blob/72e1759e0504e3a5a32587605c98e410f12ea9e7/crates/swift-bridge-ir/src/errors/parse_error.rs#L7-L22

### Get the test passing

We can get the test passing by iterating over the `attributes.args_into` that we parsed and pushing parse errors if the arg doesn't appear in the function's arguments. 
Perhaps here above the `let func = ParsedExternFn {`

https://github.com/chinedufn/swift-bridge/blob/07a15d78782dc371a7d24301ac9625124e42f8d8/crates/swift-bridge-ir/src/parse/parse_extern_mod.rs#L164-L196

NiwakaDev:
Hi, @chinedufn. Can I work on this?

Sorry...Supporting `Vec<String>` is a more difficult issue than I thought. So, I want to work on another issue first to understand the entire project.

[...]

PR #145, opened 2023-01-25
Add UI test for missing argument in args_into

NiwakaDev:
This PR addresses #11 and Introduces a compile error when using missing argument in `args_into`.

Here's an example:
```rust
#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        #[swift_bridge(args_into = (arg, arg_typo))]
        fn some_function(arg: u8);
    }
}
```

```
error: arg_typo: Unknown argument
       Could not find "arg_typo" in "fn some_function(arg : u8)".
 --> tests/ui/args-into-argument-not-found.rs:7:42
  |
7 |         #[swift_bridge(args_into = (arg, arg_typo))]
  |                                          ^^^^^^^^

```

### DIFF
--- a/crates/swift-bridge-ir/src/errors/parse_error.rs
+++ b/crates/swift-bridge-ir/src/errors/parse_error.rs
@@ -75,12 +75,9 @@ impl Into<syn::Error> for ParseError {
             ParseError::ArgsIntoArgNotFound { func, missing_arg } => Error::new_spanned(
                 missing_arg.clone(),
                 format!(
-                    r#"{}: Unknown argument
-Could not find "{}" in "{}".
-"#,
+                    r#"Argument "{}" was not found in the "fn {}(..)""#,
                     missing_arg,
-                    missing_arg,
-                    func.sig.to_token_stream().to_string()
+                    func.sig.ident.to_token_stream().to_string()
                 ),
             ),
             ParseError::AbiNameMissing {
--- a/crates/swift-bridge-ir/src/errors/parse_error.rs
+++ b/crates/swift-bridge-ir/src/errors/parse_error.rs
@@ -1,10 +1,14 @@
 use proc_macro2::Ident;
 use quote::ToTokens;
 use syn::{Error, FnArg, Item, Receiver};
-use syn::{ForeignItemType, LitStr};
+use syn::{ForeignItemFn, ForeignItemType, LitStr};
 use syn::{Token, Type};
 
 pub(crate) enum ParseError {
+    ArgsIntoArgNotFound {
+        func: ForeignItemFn,
+        missing_arg: Ident,
+    },
     /// `extern {}`
     AbiNameMissing {
         /// `extern {}`
@@ -68,6 +72,17 @@ pub(crate) enum IdentifiableParseError {
 impl Into<syn::Error> for ParseError {
     fn into(self) -> Error {
         match self {
+            ParseError::ArgsIntoArgNotFound { func, missing_arg } => Error::new_spanned(
+                missing_arg.clone(),
+                format!(
+                    r#"{}: Unknown argument
+Could not find "{}" in "{}".
+"#,
+                    missing_arg,
+                    missing_arg,
+                    func.sig.to_token_stream().to_string()
+                ),
+            ),
             ParseError::AbiNameMissing {
                 extern_token: extern_ident,
             } => Error::new_spanned(
--- a/crates/swift-bridge-ir/src/parse/parse_extern_mod.rs
+++ b/crates/swift-bridge-ir/src/parse/parse_extern_mod.rs
@@ -905,7 +905,7 @@ mod tests {
 
         let errors = parse_errors(tokens);
 
-        /// Only "bar" should be missing argument.
+        // Only "bar" should be missing argument.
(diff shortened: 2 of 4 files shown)


crates/swift-bridge-ir/src/errors/parse_error.rs
```diff
@@ -68,6 +72,17 @@ pub(crate) enum IdentifiableParseError {
 impl Into<syn::Error> for ParseError {
     fn into(self) -> Error {
         match self {
+            ParseError::ArgsIntoArgNotFound { func, missing_arg } => Error::new_spanned(
+                missing_arg.clone(),
+                format!(
+                    r#"{}: Unknown argument
```
chinedufn, line 78: Maybe:

format!(`Argument "{}" was not found in the function "{}"'s signature.", missing_arg, func.sig.ident.to_token_stream().to_string());

crates/swift-bridge-ir/src/parse/parse_extern_mod.rs
```diff
@@ -866,6 +890,25 @@ mod tests {
...
+                    #[swift_bridge(args_into = (foo, bar))]
+                    fn some_function(foo: u8);
+                }
+            }
+        };
+
+        let errors = parse_errors(tokens);
+
+        // Only "bar" should be missing argument.
+        assert_eq!(errors.len(), 1);
```
chinedufn, line 909: Can we also match on `&errors[0]` and confirm that it is an `ArgsIntoArgNotFound` error?

chinedufn: i.e. something like:

```rust
        assert_eq!(errors.len(), 1);
        match &errors[0] {
            ParseError::ArgsIntoArgNotFound {
                func: _,
                missing_arg,
            } => {
                assert_eq!(missing_arg, "bar")
            }
            _ => panic!(),
        };
```

crates/swift-bridge-ir/src/parse/parse_extern_mod.rs
```diff
@@ -180,7 +179,32 @@ impl<'a> ForeignModParser<'a> {
...
+                                    FnArg::Typed(pat_ty) => {
+                                        let func_arg_name =
+                                            pat_ty.pat.to_token_stream().to_string();
+                                        let arg_name = arg.to_token_stream().to_string();
+                                        if func_arg_name != arg_name {
+                                            count += 1;
+                                        }
+                                    }
+                                    _ => {
+                                        todo!("Add a test that hits this case")
```
chinedufn, line 196: This will panic if you use `args_into` on a method:

```rust
#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        type SomeType;
        #[swift_bridge(args_into = (foo))]
        fn some_function(&self);
    }
}
```

crates/swift-bridge-ir/src/parse/parse_extern_mod.rs
```diff
@@ -180,7 +179,32 @@ impl<'a> ForeignModParser<'a> {
                                 .push(ParseError::ArgCopyAndRefMut { arg: arg.clone() });
                         }
                     }
-
+                    if let Some(ref args) = attributes.args_into {
+                        for arg in args.iter() {
```
chinedufn, line 183: Maybe:

```rust
                    if let Some(ref args) = attributes.args_into {
                        let mut func_sig_args = HashSet::with_capacity(args.len());
                        for fn_arg in func.sig.inputs.iter() {
                            match fn_arg {
                                FnArg::Receiver(_) => {}
                                FnArg::Typed(pat_ty) => {
                                    let fn_arg_name = pat_ty.pat.to_token_stream().to_string();
                                    func_sig_args.insert(fn_arg_name);
                                }
                            }
                        }

                        for arg in args.iter() {
                            let arg_name = arg.to_token_stream().to_string();

                            if !func_sig_args.contains(&arg_name) {
                                self.errors.push(ParseError::ArgsIntoArgNotFound {
                                    func: func.clone(),
                                    missing_arg: arg.clone(),
                                })
                            }
                        }
                    }
```

crates/swift-bridge-macro/tests/ui/args-into-argument-not-found.rs
```diff
@@ -0,0 +1,18 @@
+//! # To Run
+//! cargo test -p swift-bridge-macro -- ui trybuild=args-into-argument-not-found.rs
+
+#[swift_bridge::bridge]
+mod ffi {
+    extern "Rust" {
+        #[swift_bridge(args_into = (arg, arg_typo))]
+        fn some_function(arg: u8);
+    }
```
chinedufn, line 9: Can we add a second extern "Rust" block here that uses a method?

Maybe something like:

```rust
extern "Rust" {
    type SomeType;

    #[swift_bridge(args_into = (arg, arg_typo))]
    fn some_method(&self, arg: u8);
}
```

chinedufn (commented):
Thanks a lot for working on this. Great stuff.

A few minor requests then looks good to me!

NiwakaDev:
Addressed your review!

chinedufn:
Thanks!

Issue #148, opened 2023-01-27
Optimize RustStr's Equatable implementation


NiwakaDev:
I want to address the following TODO:

https://github.com/chinedufn/swift-bridge/blob/a11b368bef4ba904c39481d735db537205de29c4/crates/swift-bridge-build/src/generate_core/string.swift#L28-L35

chinedufn:
Awesome! Let me know if you need anything.

NiwakaDev:
To address this issue, I added the following unit test:
```Swift

let hi = create_string("hi").as_str()
let he = create_string("he").as_str()
XCTAssertNotEqual(hi, he)
```

Result: 
<img width="466" alt="スクリーンショット 2023-01-29 9 45 21" src="https://user-images.githubusercontent.com/61189782/215297854-a5c99b6a-7acf-4b4c-ab2a-6a31ce3bba14.png">

`hi` and `he` seem to have the same address. It's strange. I' dont overrwrite `create_string` and `as_str` method, by the way. I just overrwrote RustStr's `==` operator.

chinedufn:
Closed via https://github.com/chinedufn/swift-bridge/pull/151

PR #151, opened 2023-01-29
Optimize `RustStr`'s `Equatable` implementation

NiwakaDev:
Addresses #148.


SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/StringTests.swift
```diff
@@ -46,5 +46,26 @@ class StringTests: XCTestCase {
...
+    func testRustStrEqualityOperator() throws {
+        XCTContext.runActivity(named: "Should be equal"){
+            _ in
+            let hello1 = create_string("hello")
+            let hello2 = create_string("hello")
+            XCTAssertEqual(hello1.as_str(), hello2.as_str())
+        }
+        XCTContext.runActivity(named: "Should not be equal"){
+            _ in
+            //Equal length
```
chinedufn, line 59: minor: The `Equal length` and `Not equal length` are on the wrong tests. Need to switch them.

crates/swift-bridge-build/src/generate_core/rust_string.c.h
```diff
@@ -15,4 +15,5 @@ void* __swift_bridge__$RustString$new(void);
 void* __swift_bridge__$RustString$new_with_str(struct RustStr str);
 uintptr_t __swift_bridge__$RustString$len(void* self);
 struct RustStr __swift_bridge__$RustString$as_str(void* self);
-struct RustStr __swift_bridge__$RustString$trim(void* self);
\ No newline at end of file
+struct RustStr __swift_bridge__$RustString$trim(void* self);
+bool __swift_bridge__equality_operator_for_RustStr(struct RustStr lhs, struct RustStr rhs);
```
chinedufn, line 19: `__swift_bridge__$RustStr$partial_eq`

src/lib.rs
```diff
@@ -92,5 +92,25 @@ pub extern "C" fn __swift_bridge__null_pointer() -> *const std::ffi::c_void {
     std::ptr::null()
 }
 
+#[no_mangle]
+#[doc(hidden)]
+pub extern "C" fn __swift_bridge__equality_operator_for_RustStr(
```
chinedufn, line 100: `__swift_bridge__RustStr__partial_eq`

src/lib.rs
```diff
@@ -92,5 +92,25 @@ pub extern "C" fn __swift_bridge__null_pointer() -> *const std::ffi::c_void {
...
+    lhs: string::RustStr,
+    rhs: string::RustStr,
+) -> bool {
+    if lhs.len != rhs.len {
+        return false;
+    }
+
+    unsafe {
```
chinedufn, line 108: Another alternative that I thought of was:

```rust
unsafe { 
    std::slice::from_raw_parts(lhs.start, lhs.len) ==
    std::slice::from_raw_parts(rhs.start, rhs.len)
}
```

I was thinking that the stdlib might have an optimization that I was unfamiliar with, but I think that desugars to https://doc.rust-lang.org/1.52.1/src/core/slice/cmp.rs.html#69-75

Which looks like the same as what you're doing. So this looks good to me.

src/lib.rs
```diff
@@ -92,5 +92,25 @@ pub extern "C" fn __swift_bridge__null_pointer() -> *const std::ffi::c_void {
     std::ptr::null()
 }
 
+#[no_mangle]
```
chinedufn, line 98: You can put this in under here https://github.com/chinedufn/swift-bridge/blob/23d60feaa0f8e92f2c854d8665868aa5573ef505/src/std_bridge/string.rs#L61-L82

chinedufn (commented):
Minor notes then looks good to me.

Great work here, thanks for figuring this out!

crates/swift-bridge-build/src/generate_core/string.swift
```diff
@@ -27,10 +27,7 @@ extension RustStr: Identifiable {
 }
 extension RustStr: Equatable {
     public static func == (lhs: RustStr, rhs: RustStr) -> Bool {
-        // TODO: Rather than compare Strings, we can avoid allocating by calling a function
-        // on the Rust side that compares the underlying byte slices.
-        return
-            lhs.toString() == rhs.toString()
+        return __swift_bridge__equality_operator_for_RustStr(lhs, rhs);
```
chinedufn, line 30: Beautiful

src/lib.rs
```diff
@@ -94,3 +94,23 @@ pub extern "C" fn __swift_bridge__null_pointer() -> *const std::ffi::c_void {
...
 #[doc(hidden)]
 pub fn assert_copy<T: Copy>() {}
+
+#[no_mangle]
+#[doc(hidden)]
+pub extern "C" fn __swift_bridge__equality_operator_for_RustStr(
+    lhs: string::RustStr,
+    rhs: string::RustStr,
+) -> bool {
+    if lhs.len != rhs.len {
```
chinedufn, line 104: Let's make this `lhs == rhs`

And then you can

```rust
impl PartialEq for RustStr { .. }
```

And handle the comparison inside of the `PartialEq` impl.

src/lib.rs
NiwakaDev: I got the link error like: 
```
Testing failed:
        Undefined symbol: ___swift_bridge__$RustStr$partial_eq
```

How does the Swift-compiler know where methods are, such as `len`, `to_string`.
https://github.com/chinedufn/swift-bridge/blob/0de0dc880709ee655aa8d625d7bbefe420456b55/src/std_bridge/string.rs#L62-L80

NiwakaDev: Seaching for `RustStr`'s methods with a tool like `grep`, I couldn't find any declarations of them in any c-header files.

chinedufn: https://github.com/chinedufn/swift-bridge/blob/e8d3859d2a68b5d5e0012185a87e88b3c811f853/crates/swift-bridge-build/src/generate_core.rs#L16-L38

https://github.com/chinedufn/swift-bridge/blob/b654a24d6d5169e0e73192ebffa556cd43044512/crates/swift-bridge-build/src/generate_core/rust_string.c.h#L17-L18

chinedufn: Oh wait, it looks like none of the `RustStr` methods like `len` or `to_string` are actually bridged.

But yeah you can declare the `___swift_bridge__$RustStr$partial_eq` inside of `rust_string.c.h`

chinedufn: Also you can leave a module comment inside of `std_bridge/string.rs` along the lines of

```rust
//! The corresponding C and Swift code can be found in
//! crates/swift-bridge-build/src/generate_core/rust_string.{c.h,swift}
```

Same idea in `std_bridge/rust_vec.rs`

And then in `std_bridge.rs`

```rust
//! The corresponding C and Swift code can be found in
//! crates/swift-bridge-build/src/generate_core/*
```

NiwakaDev: >But yeah you can declare the ___swift_bridge__$RustStr$partial_eq inside of rust_string.c.h 

I already tried to add `the ___swift_bridge__$RustStr$partial_eq` to `rust_string.c.h`. However I got the link error.

NiwakaDev: https://github.com/NiwakaDev/swift-bridge/blob/03245ec39c4cf40f7909704456617f6186f1bc50/crates/swift-bridge-build/src/generate_core/rust_string.c.h#L19

src/std_bridge/string.rs
```diff
@@ -79,4 +81,15 @@ impl RustStr {
             len: str.len(),
         }
     }
+
+    pub fn partial_eq(
```
chinedufn, line 85: Should move this into a `PartialEq` trait implementation.

```rust
impl PartialEq for RustStr {
    // ...
}
```

https://doc.rust-lang.org/std/cmp/trait.PartialEq.html

src/lib.rs
chinedufn: The build script might not be getting re-run.

Try running `touch crates/swift-integration-tests/build.rs` or otherwise modifying the `build.rs` so that it gets re-run and seeing if that works.

https://doc.rust-lang.org/cargo/reference/build-scripts.html#life-cycle-of-a-build-script

src/std_bridge/string.rs
NiwakaDev: >Should move this into a PartialEq trait implementation. 

Sorry...I know that I should do it, but I want to solve link error. So, I pushed 03245ec39c4cf40f7909704456617f6186f1bc50 to share this problem.

src/lib.rs
NiwakaDev: >Try running touch crates/swift-integration-tests/build.rs or otherwise modifying and saving

I got the link error again. 

`RustString`'s methods seem to be imported on the Swift side.
スクリーンショット 2023-01-30 1 33 35

However, `RustStr`'s methods doesn't seem to be imported on the Swift side.

スクリーンショット 2023-01-30 1 33 44

I tried to dive into this problem, and run `objdump -d libswift_integration_tests.a | grep RustStr`: 
```
0000000000000328 <__ZN12swift_bridge10std_bridge6string7RustStr10partial_eq17h39b2f23e5457666eE
```

`patial_eq` might be not `__swift_bridge__$RustStr$partial_eq` but `__ZN12swift_bridge10std_bridge6string7RustStr10partial_eq17h39b2f23e5457666eE`.

chinedufn: I just cloned your branch and I can reproduce the error. Give me a sec

chinedufn: This should work:

```rust
impl RustStr {
    pub fn len(&self) -> usize {
        self.len
    }

    // TODO: Think through these lifetimes and the implications of them...
    pub fn to_str<'a>(self) -> &'a str {
        let bytes = unsafe { std::slice::from_raw_parts(self.start, self.len) };
        std::str::from_utf8(bytes).expect("Failed to convert RustStr to &str")
    }

    pub fn to_string(self) -> String {
        self.to_str().to_string()
    }

    pub fn from_str(str: &str) -> Self {
        RustStr {
            start: str.as_ptr(),
            len: str.len(),
        }
    }
}

impl PartialEq for RustStr {
    fn eq(&self, other: &Self) -> bool {
        unsafe {
            std::slice::from_raw_parts(self.start, self.len)
                == std::slice::from_raw_parts(other.start, other.len)
        }
    }
}

#[export_name = "__swift_bridge__$RustStr$partial_eq"]
#[allow(non_snake_case)]
pub extern "C" fn __swift_bridge__RustStr_partial_eq(lhs: RustStr, rhs: RustStr) -> bool {
    lhs == rhs
}
```

NiwakaDev: I'm sorry for taking up your time. I'm going to study rust hard.

NiwakaDev:
I addressed your review. Sorry...

chinedufn:
Great work!

Issue #153, opened 2023-01-30
Support Argument-Labels


NiwakaDev:
Swift's functions have `Argument Labels` and `Parameter Names`. As far as I understand, `swift-bridge` doesn't seem to support `Argument Labels`.

Something like: 
```rust
//Rust
#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        #[swift_bridge(argument_labels=(argument_label))]
        fn get(parameter_name: i32);
    }
}

```

```Swift
//Generated Swift 

func get(argument_label parameter_name: Int){

}
```

## Resource
<a href="https://docs.swift.org/swift-book/LanguageGuide/Functions.html">Functions</a>

chinedufn:
Nice, we should totally support this.

As far as the API, what do you think about something like this:

```rust
#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        fn greet(
            #[swift_bridge(label = "firstName")] name: &str,
            age: u32
        )
    }
}
```

```swift
func greet(firstName name: RustStr, _ age: u32) {
    // ...
}
```

The argument attribute approach feels a bit cleaner than the function attribute approach.

_Now that I think of it, this is probably a better approach for things like our `get`, `get_with` and `args_into` attribute as well... But we can handle those separately._

NiwakaDev:
>The argument attribute approach makes it a bit easier tell which label applies to which argument.

Sounds good! I didn't know the argument attribute. Can I implement this?

chinedufn:
Go for it!

chinedufn:
Closed via #156

PR #156, opened 2023-02-04
Support Argument-Label for Swift

NiwakaDev:
This PR addresses #153 and introduces `#[swift_bridge(label = "...")]` to support Swift's `Argument-Label`.

Here's an example of using this:

```rust
// Rust
#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        fn test_argument_label(
            #[swift_bridge(label = "someArg")] some_arg: i32,
            another_arg: i32,
        ) -> i32;
    }
}
```

```Swift 
// Swift 
XCTAssertEqual(test_argument_label(someArg: 10, 100), 110)
```


crates/swift-bridge-ir/src/parsed_extern_fn.rs
```diff
@@ -92,6 +92,7 @@ pub(crate) struct ParsedExternFn {
     pub args_into: Option<Vec<Ident>>,
     /// Get one of the associated type's fields
     pub get_field: Option<GetField>,
+    pub argument_labels: Option<HashMap<Ident, LitStr>>,
```
chinedufn, line 95: Maybe just `HashMap<Ident, LitStr>` instead of `Option`?

crates/swift-bridge-ir/src/parse/parse_extern_mod/argument_attributes.rs
```diff
@@ -0,0 +1,85 @@
...
+impl Parse for ArgumentAttr {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        let key: Ident = input.parse()?;
+        let attribute = match key.to_string().as_str() {
+            "label" => {
+                input.parse::<Token![=]>()?;
+                let value: LitStr = input.parse()?;
+                ArgumentAttr::ArgumentLabel(value)
+            }
+            _ => {
```
chinedufn, line 41: We should add a UI test file for the unrecognized argument attribute error.

Similar to https://github.com/chinedufn/swift-bridge/blob/e0d3d62336cfb909d299ee1b75b89fc0c802b3b0/crates/swift-bridge-macro/tests/ui/unrecognized-function-attribute.rs#L1-L12

https://github.com/chinedufn/swift-bridge/blob/c75d6e744ac64e21145afbe4d251486f861e90cd/crates/swift-bridge-macro/tests/ui/unrecognized-function-attribute.stderr#L1-L5

crates/swift-bridge-ir/src/parse/parse_extern_mod/argument_attributes.rs
```diff
@@ -0,0 +1,85 @@
...
+        Ok(attribute)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::test_utils::parse_ok;
+    use quote::{format_ident, quote};
+
+    /// Verify that we can parse a function that has a argument label.
```
chinedufn, line 58: Great test.

crates/swift-bridge-ir/src/parse/parse_extern_mod/function_attributes.rs
```diff
@@ -177,13 +177,17 @@ mod tests {
 
                     #[swift_bridge(return_into)]
                     fn some_function () -> Foo;
+
+                    fn print_hello_world();
```
chinedufn, line 181: What's this? Seems like this should be removed?

crates/swift-bridge-ir/src/parse/parse_extern_mod/function_attributes.rs
```diff
@@ -177,13 +177,17 @@ mod tests {
...
                 }
             }
         };
 
         let module = parse_ok(tokens);
 
         assert!(module.functions[0].return_into);
+        assert_eq!(module.functions.len(), 2);
```
chinedufn, line 189: What's this? Seems like this should be removed?

crates/swift-bridge-ir/src/parsed_extern_fn/to_swift_func.rs
```diff
@@ -39,11 +39,26 @@ impl ParsedExternFn {
...
+                            params.push(format!("_ {}", param))
+                        }
+                    } else {
+                        let param = format!("{}: {}", arg_name, ty);
+                        params.push(format!("_ {}", param))
+                    }
                 }
             };
-
-            params.push(format!("_ {}", param))
```
chinedufn, line 46: I think we can keep the `push` on the outside and just change it to `params.push(param)`.

Then in the match block we can put the `_` or the `argument_label`.

SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/ArgumentAttributesTest.swift
```diff
@@ -0,0 +1,36 @@
...
+//
+//  Created by Niwaka on 2023/02/04.
+//
+
+import Foundation
+
+import XCTest
+@testable import SwiftRustIntegrationTestRunner
+
+/// Tests the #[swift_bridge(label = "someArg")] attribute.
```
chinedufn, line 13:
```
/// Tests argument attributes such as `#[swift_bridge(label = "someArg")]`.
```

SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/ArgumentAttributesTest.swift
```diff
@@ -0,0 +1,36 @@
...
+    override func tearDownWithError() throws {
+        // Put teardown code here. This method is called after the invocation of each test method in the class.
+    }
+
+    /// Verify that the `swift_bridge(label = "someArg")` attribute works.
+    func testArgumentLabel() throws {
+        XCTAssertEqual(test_argument_label(someArg: 10, 100), 110)
+    }
+
+    func testPerformanceExample() throws {
```
chinedufn, line 29: Can delete all of this setup and tear down and performance stuff. We never use them.

chinedufn (approved):
NNNIIIICCCEEEE!!!!

Left some minor feedback but overall this is great. Awesome work!!!

chinedufn:
One more thing. We need to add docs.

Can put them under here https://github.com/chinedufn/swift-bridge/blob/36c90e7d3aed7013575311af810f268faf263b51/book/src/bridge-module/functions/README.md#L39-L83

crates/swift-bridge-ir/src/codegen/generate_swift.rs
```diff
@@ -73,18 +73,18 @@ impl SwiftBridgeModule {
                     continue;
                 }
             }
-
             let func_definition = match function.host_lang {
                 HostLang::Rust => {
-                    gen_func_swift_calls_rust(function, &self.types, &self.swift_bridge_path)
+                    let result =
```
chinedufn, line 78: Calling this `result` makes it seem like we return a `Result`. Can we change this back?

NiwakaDev:
Addressed your review! Please let me know if there are any omissions.

chinedufn:
Thanks!

crates/swift-bridge-ir/src/parse/parse_extern_mod/function_attributes.rs
NiwakaDev: Sorry...I usually add some test-codes to understand how `swift-bridge` works. I forgot to remove this. From now on, I'll remove them before sending a PR.

Issue #162, opened 2023-02-10
Mutability safety


anforowicz:
I have a question that isn't answered by the book yet, so let me open an issue to see if the book can be tweaked to cover an answer. I see that the "Memory Safety" section of the book has a subsection about [lifetimes](https://chinedufn.github.io/swift-bridge/safety/index.html#lifetimes) - is it possible to also cover mutability safety?

In particular, if Swift passes an immutable Rust reference (`&T`) into a Rust function, then can Swift mutate the value behind such reference (e.g. if the Rust function calls back into Swift, then can the called back Swift code mutate the value)?  (Rust's language reference [says](https://doc.rust-lang.org/reference/behavior-considered-undefined.html#:~:text=Mutating%20immutable%20data.) that "mutating immutable data" triggers undefined behavior.)

It's possible that this is already taken care of and inherently safe because of Swift and Rust languages design.  If so, then I think it would be great to point that out in the book.

chinedufn:
Thanks a lot for the question and suggestion.

---

If we're only passing around a Rust `&T` then Rust's borrowing rules cannot be broken
since when the Rust `&T` shows up on the Swift side none of `T`'s mutable methods will
be available.

( See: https://github.com/chinedufn/swift-bridge/blob/bdbb028dea65148136851ff24bb8a025c5c66557/book/src/bridge-module/opaque-types/README.md?plain=1#L52-L133 )

```rust
// Rust
#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        type MyGreeter;

        fn new() -> MyGreeter;

        fn greet(&self, user: SomeSwiftUser);
        fn mutate(&mut self);
    }

    extern "Swift" {
        type SomeSwiftUser;

        fn analyze_greeter(&self, greeter: &MyGreeter);
    }
}
```

```swift
// Swift

class SomeSwiftUser {
    func analyze_greeter(greeter: MyGreeterRef) {
        // This `greeter.mutate()` would not compile since
        // it doesn't exist on `MyGreeterRef`.
        // greeter.mutate()
    }
}
```

However, it's totally possible to violate Rust's borrowing rules today.

# Violating Rust's borrowing rules

Here are the two cases that I can think of where you can violate Rust's borrowing rules.

I'm not yet sure to what extent we'll be able to mitigate these as [Swift introduces more ways to enforce ownership.](https://github.com/apple/swift/blob/main/docs/OwnershipManifesto.md).

[...]

Issue #161, opened 2023-02-10
Aliasing safety


anforowicz:
I have a question that isn't answered by the book yet, so let me open an issue to see if the book can be tweaked to cover an answer.  I see that the "Memory Safety" section of the book has a subsection about [lifetimes](https://chinedufn.github.io/swift-bridge/safety/index.html#lifetimes) - is it possible to also cover aliasing safety?

In particular, can you please clarify if/how the FFI layer enforces [aliasing safety](https://doc.rust-lang.org/reference/behavior-considered-undefined.html#:~:text=Breaking%20the%20pointer%20aliasing%20rules.)?  For example, is it possible to call the following function with the same reference passed to both arguments (AFAIK this example can be used to trigger Rust UB by making a [`cxx`](https://cxx.rs/)-mediated call from C++):

```rs
pub fn clone_from_i32(target: &mut i32, source: &i32) {
    // Rust guarantees that `&mut T` is the only, exclusive reference.
    // Having a `&mut T` that is aliased by other references leads to
    // Undefined Behavior in Rust.
    let target_ptr: *mut i32 = target;
    let source_ptr: *const i32 = source;
    assert_ne!(target_ptr.cast_const(), source_ptr);

    target.clone_from(source);
}
```

If it is not possible to cause an aliasing violation, then can you clarify how this is achieved?  Maybe this aspect of safety is already guaranteed by the [runtime exclusivity checks in Swift](https://www.swift.org/blog/swift-5-exclusivity/)?

chinedufn:
At a high level this folds under #162 in that [Rust's borrowing rules](https://doc.rust-lang.org/book/ch04-02-references-and-borrowing.html#the-rules-of-references) are not currently fully enforced across FFI boundaries.

`swift-bridge` does not currently guarantee aliasing safety. For example, the following is currently possible:

```rust
// Rust

#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        type RustType;

        #[swift_bridge(init)]
        fn new() -> RustType;

        fn combine(a: &mut RustType, b: &RustType);
    }
}
```

```swift
// Swift

let rustType = RustType()

// Violates Rust's borrowing rules.
rustType.combine(rustType)
```

---

Whether or not Swift's upcoming ownership related features will allow us to better enforce Rust's borrowing rules across the FFI boundary is an open question that needs more research.

---

[...]

PR #164, opened 2023-02-11
Improve memory safety documentation

chinedufn:
This commit adds documentation that explains that it is possible to
trigger undefined behavior by violating Rust's borrowing rules using
Swift.

Related:
- https://github.com/chinedufn/swift-bridge/issues/161
- https://github.com/chinedufn/swift-bridge/issues/162

### DIFF
--- a/book/src/safety/README.md
+++ b/book/src/safety/README.md
@@ -4,18 +4,24 @@
 
 ## Type Safety
 
-The code that `swift-bridge` is type safe on both the Rust and Swift side,
-so all of your interfacing between the two languages is type safe.
+All of the Rust and Swift FFI code that `swift-bridge` generates
+for you is type safe.
 
 ## Memory Safety
 
-Needing to worry about memory safety should be very uncommon when using `swift-bridge`.
+You can ensure the memory safety of your Rust and Swift projects by following these rules:
 
-There are two known situations that can lead to unsafe memory access.
+- In Swift, never use a reference to a Rust type after its lifetime.
 
-#### Lifetimes
+- In Swift, pass a mutable reference to Rust that will live alongside another active reference to that type.
 
-It's possible to pass a reference from `Rust` -> `Swift` and then have `Swift` make use of that
+- In Swift, never use a type after passing ownership to Rust.
+
+Let's look at some examples of code that violates these rules:
+
+### Never use a reference after its lifetime
+
+It is possible to pass a reference from `Rust` -> `Swift` and then have `Swift` make use of that
 reference after it is no longer safe to do so.
 
 ```rust
@@ -34,6 +40,8 @@ mod ffi {
 ```
 
 ```swift
+// Swift
+
 let someType = SomeType()
 
 let name: RustStr = someType.name()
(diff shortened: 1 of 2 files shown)


chinedufn:
Rendered https://github.com/chinedufn/swift-bridge/blob/031fe4c8987d4dd1b1e653f1af5495229ee43f34/book/src/safety/README.md#breaking-rusts-borrowing-rules

chinedufn:
@anforowicz I'm replying to your comments in #161 and #162 here.

---

I'm thinking that the main takeaway that we want the reader to have is:

> If I have a mutable reference AND one or more other references to the same value, I'm in the danger zone and need to be very careful.

If we are to assume that the majority of users will be writing safe Swift and safe Rust code, then we should expect that the majority of users are used to never needing to think deeply about memory safety. 

So, they just need a high level sense of what not to do and why not to do it.

---

I'm imagining that the main thing that the FFI auditor wants to know is:

> Is the Rust/Swift FFI fully safe? If not, when is it unsafe? What are all the ways to trigger undefined behavior?

So, when it comes to memory safety, I think that we want the security auditor to have the same main takeaway as the regular user (anywhere that there is a mutable reference and at least one other reference should be audited).

---

The above led me to opt for a single simple section on Rust's borrowing rules.

Please let me know if you think that either the regular user and/or the FFI auditor is underserved by this. 

If so, we can make some changes. If not, we can land this as one section for now.

book/src/safety/README.md
```diff
@@ -4,18 +4,22 @@
...
-The code that `swift-bridge` is type safe on both the Rust and Swift side,
-so all of your interfacing between the two languages is type safe.
+All of the Rust and Swift FFI code that `swift-bridge` generates
+for you is type safe.
 
 ## Memory Safety
 
-Needing to worry about memory safety should be very uncommon when using `swift-bridge`.
+`swift-bridge` guarantees that the Rust side of your FFI boundary
+is memory safe.
```
anforowicz, line 13: First let me try to explain what I think can be improved in the snippet above.

> the Rust side of your FFI boundary is memory safe

This seems to be a bit vague.
- One counter-example is that safe Rust can end up dereferencing an already freed memory (e.g. if Rust gets `&'static T` from Swift but Swift frees that).
- I guess that the sentence above tries to say that safe Rust is not the **root** cause of the UaF (e.g. tries to say that the Swift side of FFI is at fault).

> `swift-bridge` guarantees

This may be misunderstood / this seems to overpromise?

Maybe you would consider rewriting to something like:

-------------------

The memory safety of a Rust program depends on following a few rules in `unsafe` blocks.  A program using a mix of Swift and Rust needs to ensure that  Swift code follows similar rules.  Most of the rules are related to references:
- When Swift receives a reference from Rust, one needs to ensure that:
    -  Swift doesn't dereference the value outside of the reference lifetime
- When Swift passes a reference to Rust, one needs to ensure that:
    - Rust lifetime expectations are met
    - Mutable Rust references are exclusive (i.e. there are no other Rust references to the same memory)
    - Swift doesn't mutate data behind immutable Rust references

anforowicz:
I think this PR is a strict improvement, so even though I'd like to propose some tweaks, I think it's best to land it and then continue iterating if needed/desired.  I agree with your main point that the guidance should be easy to absorb (while keeping it accurate as much as possible).  I tried to proposed a tweak of the introductory paragraph in a separate comment.

book/src/safety/README.md
chinedufn: Super clean!! Thanks!!

I'll apply your feedback pretty much as is.

chinedufn:
Agreed. Will land this at some point this week. Thanks for the terrific feedback.

anforowicz:
Thanks!  Feel free to mark the 2 issues I've opened as resolved after this PR lands.  (When opening the issues I mostly wanted to see explicit mention of aliasing and mutability rules [and whether they are the responsibility of the author of Swift code].)

PR #169, opened 2023-02-13
Fix to make incorrect return type into error

### LINKED ISSUE #12: Fix bug when defining `-> OpaqueRustType` even though the real function is `-> &OpaqueRustType`
Right now this compiles and it shouldn't

```rust
struct SomeType;

#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        // The real implementation below returns &SomeType
        fn foo() -> SomeType;
    }
}

fn foo () -> &SomeType {
    unimplemented!()
}
```

---

This also compiles, even though it shouldn't.

```rust
struct SomeType;

#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        // The real implementation below returns Option<SomeType>
        fn foo() -> SomeType;
    }
}

fn foo () -> Option<SomeType> {
    unimplemented!()
}
```

### UI Test

We can add these signatures to an `incorrect_return_type.rs` UI test, similar to the https://github.com/chinedufn/swift-bridge/blob/a11b368bef4ba904c39481d735db537205de29c4/crates/swift-bridge-macro/tests/ui/incorrect-argument-type.rs#L1-L24 ui tests and then fix the code generation so that we get compile time errors for these mistakes. 

### Codegen Tests

Change the return opaque Rust type codegen test to use an explicit type annotation

https://github.com/chinedufn/swift-bridge/blob/607e30579ac4de1fab5ff66838253cf3d07ddfbb/crates/swift-bridge-ir/src/codegen/codegen_tests/extern_rust_function_opaque_rust_type_return_codegen_tests.rs#L22-L29

Something like:

```rust
Box::into_raw(Box::new({let val: super::SomeType = super::some_function(); val})) as *mut super::SomeType
```

Same idea for the reference and reference mut codegen tests

https://github.com/chinedufn/swift-bridge/blob/607e30579ac4de1fab5ff66838253cf3d07ddfbb/crates/swift-bridge-ir/src/codegen/codegen_tests/extern_rust_function_opaque_rust_type_return_codegen_tests.rs#L78-L85

https://github.com/chinedufn/swift-bridge/blob/607e30579ac4de1fab5ff66838253cf3d07ddfbb/crates/swift-bridge-ir/src/codegen/codegen_tests/extern_rust_function_opaque_rust_type_return_codegen_tests.rs#L134-L141

### Get tests passing

You can get tests passing by modifying the `BridgeableType` implementation in this file https://github.com/chinedufn/swift-bridge/blob/54b9c09337701dd7580010d1d5ef6527b855ee98/crates/swift-bridge-ir/src/bridged_type/bridged_opaque_type.rs#L206

MikuroXina:
Fixes #12. However the error diagnosis may possibly unfriendly.

### DIFF
--- a/crates/swift-bridge-macro/tests/ui/incorrect-return-type.stderr
+++ b/crates/swift-bridge-macro/tests/ui/incorrect-return-type.stderr
@@ -1,0 +1,23 @@
+error[E0308]: mismatched types
+ --> tests/ui/incorrect-return-type.rs:6:1
+  |
+6 | #[swift_bridge::bridge]
+  | ^^^^^^^^^^^^^^^^^^^^^^^
+  | |
+  | expected struct `SomeType`, found `&SomeType`
+  | expected due to this
+  |
+  = note: this error originates in the attribute macro `swift_bridge::bridge` (in Nightly builds, run with -Z macro-backtrace for more info)
+
+error[E0308]: mismatched types
+ --> tests/ui/incorrect-return-type.rs:6:1
+  |
+6 | #[swift_bridge::bridge]
+  | ^^^^^^^^^^^^^^^^^^^^^^^
+  | |
+  | expected struct `SomeType`, found enum `Option`
+  | expected due to this
+  |
+  = note: expected struct `SomeType`
+               found enum `Option<SomeType>`
+  = note: this error originates in the attribute macro `swift_bridge::bridge` (in Nightly builds, run with -Z macro-backtrace for more info)
--- a/crates/swift-bridge-ir/src/bridged_type/bridged_opaque_type.rs
+++ b/crates/swift-bridge-ir/src/bridged_type/bridged_opaque_type.rs
@@ -211,7 +211,10 @@ impl BridgeableType for OpaqueForeignType {
                     .generics
                     .angle_bracketed_concrete_generics_tokens(types);
                 quote! {
-                    Box::into_raw(Box::new(#expression)) as *mut super::#ty_name #generics
+                    Box::into_raw(Box::new({
+                        let val: super::#ty_name #generics = #expression;
+                        val
+                    })) as *mut super::#ty_name #generics
                 }
             }
         } else {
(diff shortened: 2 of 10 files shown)


crates/swift-bridge-macro/tests/ui/incorrect-return-type.stderr
```diff
@@ -0,0 +1,23 @@
+error[E0308]: mismatched types
+ --> tests/ui/incorrect-return-type.rs:6:1
+  |
+6 | #[swift_bridge::bridge]
```
chinedufn, line 4: We want to use the function's return type's span such that in

```rust
fn foo () -> SomeType;
```

the `SomeType` would get underlined.

---

To accomplish this you can change this https://github.com/chinedufn/swift-bridge/blob/54b9c09337701dd7580010d1d5ef6527b855ee98/crates/swift-bridge-ir/src/parsed_extern_fn/to_extern_c_fn.rs#L36

to something like:

```rust
        let ret_span = self.sig.output.span();
        let ret = self.rust_fn_sig_return_tokens(swift_bridge_path, types);
        let ret = quote_spanned! {ret_span=> #ret};
```

chinedufn (commented):
Thanks for your first contribution!

Left one minor request then after that we can merge this.

crates/swift-bridge-macro/tests/ui/incorrect-return-type.stderr
MikuroXina: I have tried that spanning method, but it does not produce the spanned error output. I will try to investigate the reason why.

chinedufn:
@MikuroXina I merged your PR as is, thanks for the contribution!

I added a commit message to illustrate the change. We try to explain every commit using a thorough example.

I've also updated your PR body with an example.

chinedufn:
Sorry for pointing you in the wrong direction.

I made a change to improve the compile time error message here https://github.com/chinedufn/swift-bridge/commit/1b33e95b9c5b2d87ca2485aa972dcbb6f971ef18

Here's the new message:

https://github.com/chinedufn/swift-bridge/blob/1b33e95b9c5b2d87ca2485aa972dcbb6f971ef18/crates/swift-bridge-macro/tests/ui/incorrect-return-type.stderr#L1-L33

chinedufn:
Closed #12

Issue #160, opened 2023-02-10
Support transparent enums with associated data


chinedufn:
Right now we support enums with unit variants, but we
have not yet implemented support for enums that have data.

```rust
// This is supported
enum ThisWorks {
    VariantA,
    VariantB,
}

// This is not yet supported
enum ThisDoesNotWorkYet {
    A(String),
    B {
        field: u32
    },
    C(Vec<u8>, i32),
    D {
        field1: u32,
        field2: SomeOpaqueRustType
    }
}
```

We'll need to add support for named and unnamed associaed data.

We do not need to support all of this in one pull request.

## Support Checklist

The following should be verified in tests:

- [x] Enum variant with one unnamed field
  - https://github.com/chinedufn/swift-bridge/pull/172

- [x] Enum variant with 2 unnamed fields
  - https://github.com/chinedufn/swift-bridge/pull/172

- [x] Enum variant with 1 named field
  - https://github.com/chinedufn/swift-bridge/pull/175

- [x] Enum variant with 2 unnamed field
  - https://github.com/chinedufn/swift-bridge/pull/175

### Contributing Guide

https://chinedufn.github.io/swift-bridge/contributing/adding-support-for-a-signature/index.html

### Transparent struct codegen and integration tests

Bridging enum variants should work similarly to bridging structs, so we may want to look over
the struct codegen and integration tests.

https://github.com/chinedufn/swift-bridge/blob/c1fd7177b5d3a8918e34b6bc96ae29437cfc1487/crates/swift-bridge-ir/src/codegen/codegen_tests/transparent_struct_codegen_tests.rs#L1-L704

https://github.com/chinedufn/swift-bridge/blob/b74432885120fbc6025349d8694505ec3b950918/crates/swift-integration-tests/src/shared_types/shared_struct.rs#L1-L80

https://github.com/chinedufn/swift-bridge/blob/b74432885120fbc6025349d8694505ec3b950918/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/SharedStructTests.swift#L11-L45

### Enum codegen and integration tests

Here are the existing enum codegen and integration tests:

https://github.com/chinedufn/swift-bridge/blob/819e39160ec5b38dce2e60e2d39d078617ad158d/crates/swift-bridge-ir/src/codegen/codegen_tests/transparent_enum_codegen_tests.rs#L1-L348

[...]

PR #172, opened 2023-02-18
Support enums with unnamed data

NiwakaDev:
This PR addresses #160 and adds support for enums with unnamed data. This PR doesn't implement enums with named data.

```rust
//Rust
#[swift_bridge::bridge]
mod ffi {
    enum EnumWithUnnamedData {
        Variant1(String, u32),
        Variant2(i32, u8),
        Variant3,
    }
}
```

```Swift
//Swift
let enumWithUnnamedData = EnumWithUnnamedData.Variant1(create_string("hello"), 0)
switch enumWithUnnamedData {
case .Variant1(let rustString, let valueUInt32):
    XCTAssertEqual(rustString.toString(), "hello")
    XCTAssertEqual(valueUInt32, 0)
default:
    XCTFail()
}

```

### DIFF
--- a/crates/swift-bridge-ir/src/bridged_type/shared_enum.rs
+++ b/crates/swift-bridge-ir/src/bridged_type/shared_enum.rs
@@ -7,6 +7,8 @@ use syn::LitStr;
 mod enum_variant;
 pub(crate) use self::enum_variant::EnumVariant;
 
+use super::{shared_struct::StructField, StructFields};
+
 #[derive(Clone)]
 pub(crate) struct SharedEnum {
     pub name: Ident,
@@ -53,13 +55,18 @@ impl SharedEnum {
         quote! { #name }
     }
 
+    /// __swift_bridge__$SomeEnumFields
     pub fn ffi_union_name_string(&self) -> String {
         format!("{}Fields", self.ffi_name_string())
     }
 
+    /// { __swift_bridge__$SomeEnum$FieldOfVariant1 Variant1; __swift_bridge__$SomeEnum$FieldOfVariant2 Variant2;}
     pub fn ffi_union_field_names_string(&self) -> String {
         let mut union_fields = "{".to_string();
         for variant in self.variants.iter() {
+            if let StructFields::Unit = variant.fields {
+                continue;
+            }
             let union_field = format!(
                 " {} {};",
                 variant.union_name_string(&self.ffi_name_string()),
--- a/crates/swift-bridge-ir/src/bridged_type/shared_enum.rs
+++ b/crates/swift-bridge-ir/src/bridged_type/shared_enum.rs
@@ -7,7 +7,7 @@ use syn::LitStr;
 mod enum_variant;
 pub(crate) use self::enum_variant::EnumVariant;
 
-use super::{shared_struct::StructField, StructFields};
+use super::StructFields;
 
 #[derive(Clone)]
 pub(crate) struct SharedEnum {
--- a/crates/swift-bridge-ir/src/bridged_type/shared_enum.rs
+++ b/crates/swift-bridge-ir/src/bridged_type/shared_enum.rs
@@ -53,6 +53,24 @@ impl SharedEnum {
         quote! { #name }
     }
 
+    pub fn ffi_union_name_string(&self) -> String {
+        format!("{}Fields", self.ffi_name_string())
+    }
+
+    pub fn ffi_union_field_names_string(&self) -> String {
+        let mut union_fields = "{".to_string();
+        for variant in self.variants.iter() {
+            let union_field = format!(
+                " {} {};",
+                variant.union_name_string(&self.ffi_name_string()),
+                variant.name
+            );
+            union_fields += &union_field;
+        }
+        union_fields += "}";
+        union_fields
+    }
+
     /// __swift_bridge__$Option$SomeEnum
     pub fn ffi_option_name_string(&self) -> String {
         format!(
--- a/crates/swift-bridge-ir/src/bridged_type/shared_enum/enum_variant.rs
+++ b/crates/swift-bridge-ir/src/bridged_type/shared_enum/enum_variant.rs
@@ -1,13 +1,227 @@
-use crate::bridged_type::StructFields;
+use crate::bridged_type::{BridgedType, StructFields, TypePosition};
+use crate::parse::TypeDeclarations;
 use proc_macro2::Ident;
+use proc_macro2::TokenStream;
+use quote::quote;
 use std::fmt::{Debug, Formatter};
+use syn::spanned::Spanned;
+use syn::Path;
 
 #[derive(Clone)]
 pub(crate) struct EnumVariant {
     pub name: Ident,
-    // Will be used in a future commit.
     #[allow(unused)]
     pub fields: StructFields,
+}
+
+impl EnumVariant {
+    pub(crate) fn convert_rust_expression_to_ffi_repr(
+        &self,
+        types: &TypeDeclarations,
+        swift_bridge_path: &Path,
+        enum_name: &Ident,
+        ffi_enum_name: &Ident,
+    ) -> TokenStream {
+        let variant_name = &self.name;
+        let rust_fields: Vec<TokenStream> = self
+            .fields
+            .normalized_fields()
+            .iter()
+            .map(|norm_field| norm_field.to_enum_field(&quote! {value}))
+            .collect();
+        let converted_fields: Vec<TokenStream> = self
+            .fields
+            .normalized_fields()
+            .iter()
+            .map(|norm_field| {
+                let maybe_name_and_colon = norm_field.maybe_name_and_colon();
+                let access_field = norm_field.to_enum_field(&quote! {value});
+                let ty = BridgedType::new_with_type(&norm_field.ty, types).unwrap();
+                let converted_field = ty.convert_rust_expression_to_ffi_type(
+                    &access_field,
+                    swift_bridge_path,
+                    types,
+                    norm_field.ty.span(),
+                );
+
+                quote! {
+                    #maybe_name_and_colon #converted_field
+                }
+            })
+            .collect();
+
+        let rust_fields = self.wrap_fields(&rust_fields);
+        let converted_fields = self.wrap_fields(&converted_fields);
+
+        if self.fields.is_empty() {
+            quote! {
+                #enum_name :: #variant_name => #ffi_enum_name :: #variant_name (123)
+            }
+        } else {
+            quote! {
+                #enum_name :: #variant_name #rust_fields => #ffi_enum_name :: #variant_name #converted_fields
+            }
+        }
+    }
+    pub(crate) fn convert_ffi_repr_to_rust(
+        &self,
+        swift_bridge_path: &Path,
+        types: &TypeDeclarations,
+        enum_name: &Ident,
+        ffi_enum_name: &Ident,
+    ) -> TokenStream {
+        let variant_name = &self.name;
+        let ffi_fields: Vec<TokenStream> = self
+            .fields
+            .normalized_fields()
+            .iter()
+            .map(|norm_field| norm_field.to_enum_field(&quote! {value}))
+            .collect();
+        let converted_fields: Vec<TokenStream> = self
+            .fields
+            .normalized_fields()
+            .iter()
+            .map(|norm_field| {
+                let maybe_name_and_colon = norm_field.maybe_name_and_colon();
+                let access_field = norm_field.to_enum_field(&quote!(value));
+
+                let ty = BridgedType::new_with_type(&norm_field.ty, types).unwrap();
+                let converted_field = ty.convert_ffi_expression_to_rust_type(
+                    &access_field,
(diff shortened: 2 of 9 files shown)


crates/swift-bridge-ir/src/bridged_type/shared_enum.rs
```diff
@@ -53,6 +53,24 @@ impl SharedEnum {
         quote! { #name }
     }
 
+    pub fn ffi_union_name_string(&self) -> String {
+        format!("{}Fields", self.ffi_name_string())
+    }
+
+    pub fn ffi_union_field_names_string(&self) -> String {
```
chinedufn, line 60: Can you add a doc comment with an example of a string that this might create? 
(Like the other methods in this file)

This makes it easier for new contributors to more quickly understand what these utility methods are doing.

crates/swift-bridge-ir/src/bridged_type/shared_enum/enum_variant.rs
```diff
@@ -1,15 +1,229 @@
...
+            StructFields::Named(_) => {
+                todo!();
+            }
+            StructFields::Unnamed(_) => {
+                quote! {
+                    ( #(#fields),* )
+                }
+            }
+            StructFields::Unit => {
+                debug_assert_eq!(fields.len(), 0);
```
chinedufn, line 216: I'd rather not start introducing debug assertions unless they're adding clear value.

I would consider a debug assertion valuable if it was used to uphold a safety guarantee. 
That doesn't seem like the case here.
If someone sees a debug assertion in our codebase it should mean that, despite having a test suite, there is so much danger in things going wrong that we are also adding in this (debug) runtime check of our invariants.

Since we have a lot of testing around this flow, I think we can delete this assertion.

crates/swift-bridge-ir/src/codegen/codegen_tests/transparent_enum_codegen_tests.rs
```diff
@@ -30,8 +30,8 @@ mod generates_enum_to_and_from_ffi_conversions_no_data {
             #[repr(C)]
             #[doc(hidden)]
             pub enum __swift_bridge__SomeEnum {
-                Variant1,
-                Variant2
+                Variant1(u8),
```
chinedufn, line 33: We don't need the `u8` since `#[repr(C)]` enums are FFI safe.

crates/swift-bridge-ir/src/codegen/codegen_tests/transparent_enum_codegen_tests.rs
```diff
@@ -346,3 +356,243 @@ mod shared_enum_swift_name_attribute {
...
+union __swift_bridge__$SomeEnumFields { __swift_bridge__$SomeEnum$FieldOfVariant1 Variant1; __swift_bridge__$SomeEnum$FieldOfVariant2 Variant2;};
+typedef enum __swift_bridge__$SomeEnumTag { __swift_bridge__$SomeEnum$Variant1, __swift_bridge__$SomeEnum$Variant2, } __swift_bridge__$SomeEnumTag;
+typedef struct __swift_bridge__$SomeEnum { __swift_bridge__$SomeEnumTag tag; union __swift_bridge__$SomeEnumFields payload;} __swift_bridge__$SomeEnum;
+typedef struct __swift_bridge__$Option$SomeEnum { bool is_some; __swift_bridge__$SomeEnum val; } __swift_bridge__$Option$SomeEnum;
+"#,
+        )
+    }
+
+    #[test]
+    fn generates_enum_to_and_from_ffi_conversions_no_data() {
```
chinedufn, line 469: generates_enum_to_and_from_ffi_conversions_one_unnamed_data_and_no_fields

crates/swift-bridge-ir/src/codegen/codegen_tests/transparent_enum_codegen_tests.rs
```diff
@@ -346,3 +356,243 @@ mod shared_enum_swift_name_attribute {
...
+union __swift_bridge__$SomeEnumFields { __swift_bridge__$SomeEnum$FieldOfA A; __swift_bridge__$SomeEnum$FieldOfB B;};
+typedef enum __swift_bridge__$SomeEnumTag { __swift_bridge__$SomeEnum$A, __swift_bridge__$SomeEnum$B, } __swift_bridge__$SomeEnumTag;
+typedef struct __swift_bridge__$SomeEnum { __swift_bridge__$SomeEnumTag tag; union __swift_bridge__$SomeEnumFields payload;} __swift_bridge__$SomeEnum;
+typedef struct __swift_bridge__$Option$SomeEnum { bool is_some; __swift_bridge__$SomeEnum val; } __swift_bridge__$Option$SomeEnum;
+"#,
+        )
+    }
+
+    #[test]
+    fn generates_enum_to_and_from_ffi_conversions_no_data() {
```
chinedufn, line 589: generates_enum_to_and_from_ffi_conversions_unnamed_data_case_and_two_unnamed_data

crates/swift-bridge-ir/src/codegen/generate_c_header.rs
```diff
@@ -141,24 +143,61 @@ typedef struct {option_ffi_name} {{ bool is_some; {ffi_name} val; }} {option_ffi
                         let mut variants = "".to_string();
 
                         for variant in ty_enum.variants.iter() {
-                            let v = format!("{}${}, ", ffi_name, variant.name);
-                            variants += &v;
+                            let variant = format!("{}${}, ", ffi_name, variant.name);
```
chinedufn, line 146: Nice

crates/swift-bridge-ir/src/codegen/generate_c_header.rs
```diff
@@ -141,24 +143,61 @@ typedef struct {option_ffi_name} {{ bool is_some; {ffi_name} val; }} {option_ffi
...
+                                    let variant_field = format!("typedef struct {ffi_name}$FieldOf{variant_name} {{uint8_t _private;}} {ffi_name}$FieldOf{variant_name};", ffi_name = ffi_name, variant_name = variant.name);
+                                    variant_fields += &variant_field;
+                                    variant_fields += "\n";
+                                }
+                            }
+                        }
                         let enum_decl = format!(
-                            r#"typedef enum {ffi_tag_name} {{ {variants}}} {ffi_tag_name};
-typedef struct {ffi_name} {{ {ffi_tag_name} tag; }} {ffi_name};
+                            r#"#include <stdint.h>
```
chinedufn, line 188: We don't need this unconditional `stdint.h`.

Instead we can insert includes as needed like we do for structs https://github.com/chinedufn/swift-bridge/blob/0f436bc8f55f68cf7d023e5e04e7898c428776bb/crates/swift-bridge-ir/src/codegen/generate_c_header.rs#L83-L89

chinedufn (commented):
This is impressive. Very solid work. 
Thanks for diving into this and figuring out all of the FFI bits and pieces.
Thanks for looking into unions and deciding that they were a good path forward for our enum FFI representation.

This is great.

Left some minor feedback then after that this looks good to me.

crates/swift-bridge-ir/src/codegen/codegen_tests/transparent_enum_codegen_tests.rs
NiwakaDev: If Rust's `enum` type has no fields, the corresponding `union` and the corresponding `struct` in C is generated as something like this:

```rust
#[repr(C)]
pub enum MyEnum {
    A(i32),
    B,
}
```

```C
//C header 

typedef struct FieldsOfA {int32_t _0;} FieldsOfA;
typedef struct FieldsOfB {} FieldsOfB;  // This struct is empty.

union MyEnumFields {
    FieldsOfA A;
    FieldsOfB B;
};

typedef enum MyEnumTag { _A, _B, _C,} MyEnumTag;

typedef struct MyEnum {
    enum MyEnumTag tag;
    union MyEnumFields payload;
} MyEnum;

```


`FieldsOfB` is empty. Is this safety?

chinedufn: Thanks for illustrating.

We don't need `FieldsOfB` in the `C` header union at all.

Look at `MyEnum::D` here https://rust-lang.github.io/unsafe-code-guidelines/layout/enums.html#explicit-repr-annotation-with-c-compatibility

If you look at the corresponding C header file the `C` union doesn't have the `D` variant.

---

So, we can remove the `u8`, and not even generate the `FieldsOfB` at all.

For enums where where none of the fields are data-carrying we shouldn't generate the `union` and instead use the C header format

https://github.com/chinedufn/swift-bridge/blob/master/crates/swift-bridge-ir/src/codegen/codegen_tests/transparent_enum_codegen_tests.rs#L102-L103

---

Of course all of this is probably fairly minor since we'd only be saving a single byte of memory per enum... 
But I'd rather just do it right the first time.

NiwakaDev: > If you look at the corresponding C header file the C union doesn't have the D variant.

I guess that we have to set some value to the `payload` variable. 

Considering the example above, do we set `FieldsOfA` to `palyload` when `tag` is `B`?

Something like: 
```Swift 
MyEnum(tag: B, payload:  MyEnumFields(A: FieldsOfA(_0: value)))
```

NiwakaDev: As far as I understand, the `unit` type in Rust does not exist in the corresponding type in Swift. I checked [How Swift imports C APIs](https://github.com/apple/swift/blob/main/docs/HowSwiftImportsCAPIs.md#how-swift-imports-c-apis).

chinedufn: It looks like we can use the default initializer. Explained here https://developer.apple.com/documentation/swift/using-imported-c-structs-and-unions-in-swift .

So, we want to generate something like this ` { var val = __swift_bridge__$EnumWithUnnamedData(); val.tag = __swift_bridge__$EnumWithUnnamedData$Variant3; return val } ()`

Here I've commented out the old return and added a new one that would work without specifying a payload:

```swift
public enum EnumWithUnnamedData {
    case Variant1(RustString, UInt32)
    case Variant2(Int32, UInt8)
    case Variant3
}
extension EnumWithUnnamedData {
    func intoFfiRepr() -> __swift_bridge__$EnumWithUnnamedData {
        switch self {
            case EnumWithUnnamedData.Variant1(let value_0, let value_1):
                return __swift_bridge__$EnumWithUnnamedData(tag: __swift_bridge__$EnumWithUnnamedData$Variant1, payload: __swift_bridge__$EnumWithUnnamedDataFields(Variant1: __swift_bridge__$EnumWithUnnamedData$FieldOfVariant1(_0: { let rustString = value_0.intoRustString(); rustString.isOwned = false; return rustString.ptr }(), _1: value_1)))
            case EnumWithUnnamedData.Variant2(let value_0, let value_1):
                return __swift_bridge__$EnumWithUnnamedData(tag: __swift_bridge__$EnumWithUnnamedData$Variant2, payload: __swift_bridge__$EnumWithUnnamedDataFields(Variant2: __swift_bridge__$EnumWithUnnamedData$FieldOfVariant2(_0: value_0, _1: value_1)))
            case EnumWithUnnamedData.Variant3:
//                return __swift_bridge__$EnumWithUnnamedData(tag: __swift_bridge__$EnumWithUnnamedData$Variant3, payload: __swift_bridge__$EnumWithUnnamedDataFields(Variant3: __swift_bridge__$EnumWithUnnamedData$FieldOfVariant3(_private: 123)))
            return { var val = __swift_bridge__$EnumWithUnnamedData(); val.tag = __swift_bridge__$EnumWithUnnamedData$Variant3; return val } ()
        }
    }
}
```

NiwakaDev: Thanks you!!

chinedufn: Yup! And then if none of the variants carry data the codegen should look like it did before this PR.

https://github.com/chinedufn/swift-bridge/blob/master/crates/swift-bridge-ir/src/codegen/codegen_tests/transparent_enum_codegen_tests.rs#L65-L81

NiwakaDev:
Addressed your review. However, the implementation might be naive... Please let me know if there are any omissions.

chinedufn (approved):
Thanks!

chinedufn:
Made one small change https://github.com/chinedufn/swift-bridge/commit/fe8f5216e6dc7c27cf87c3a72d687e3c3703d326

NiwakaDev:
I didn't know Iterator's `all` method. Thank you!

https://github.com/chinedufn/swift-bridge/blob/fe8f5216e6dc7c27cf87c3a72d687e3c3703d326/crates/swift-bridge-ir/src/bridged_type/shared_enum.rs#L106

PR #175, opened 2023-02-21
Support enums with named data

NiwakaDev:
This PR addresses #160 and adds support for enums with named data. 

```rust
#[swift_bridge::bridge]
mod ffi {
    enum EnumWithNamedData {
        Variant1 { hello: String, data_u8: u8 },
        Variant2 { data_i32: i32 },
        Variant3,
    }

    extern "Rust" {
        fn reflect_enum_with_named_data(arg: EnumWithNamedData) -> EnumWithNamedData;
    }
}
```

```Swift
func testEnumWithNamedData() {
    let enumWithNamedData1 = EnumWithNamedData.Variant1(hello: create_string("hello"), data_u8: 123)
    switch reflect_enum_with_named_data(enumWithNamedData1) {
    case .Variant1(let hello, let dataU8):
        XCTAssertEqual(hello.toString(), "hello")
        XCTAssertEqual(dataU8, 123)
    default:
        XCTFail()
    }
}
```

### DIFF
--- a/crates/swift-bridge-ir/src/bridged_type/shared_struct/struct_field/normalized_field.rs
+++ b/crates/swift-bridge-ir/src/bridged_type/shared_struct/struct_field/normalized_field.rs
@@ -38,6 +38,20 @@ impl NormalizedStructField {
         }
     }
 
+    /// Used when we want to avoid putting spaces at all between the field name and the colon.
+    /// // Example:
+    /// description: String // no spaces between "description" and the colon.
+    pub fn maybe_name_and_colon_string(&self) -> String {
+        match &self.accessor {
+            NormalizedStructFieldAccessor::Named(name) => {
+                format!("{}: ", name.to_string())
+            }
+            NormalizedStructFieldAccessor::Unnamed(_idx) => {
+                format!("")
+            }
+        }
+    }
+
     /// Access a struct's field
     ///
     /// // Example named field access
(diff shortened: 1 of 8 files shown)


chinedufn:
Mind surrounding your Rust example in a bridge module?

```rust
#[swift_bridge::bridge]
mod ffi {
    // ... 
}
```

crates/swift-bridge-ir/src/bridged_type/shared_struct/struct_field/normalized_field.rs
```diff
@@ -38,6 +38,20 @@ impl NormalizedStructField {
         }
     }
 
+    /// Used when we want to avoid putting spaces at all between the field name and the colon.
```
chinedufn, line 41: I think it would be clearer to do all of the field setting here instead of doing half of it here an half of it at the call site. 

Maybe something like:

```rust
/// Create a string for setting the value of a struct's field.
///
/// Example if named field -> "field_name: someValue".
/// Example if unnamed field -> "someValue".
struct_field_setter_string(&self, value: String) -> String {
    match &self.accessor {
        NormalizedStructFieldAccessor::Named(name) => {
            format!("{}: {}", name.to_string(), value)
        }
        NormalizedStructFieldAccessor::Unnamed(_idx) => {
            value
        }
    }
}

```

chinedufn (approved):
Great work. One minor piece of feedback but otherwise looks good to me.

Thanks!

chinedufn:
Great stuff. Thanks!

Issue #177, opened 2023-02-24
Support for returning Result<(), OpaqueType>


jfaust:
Currently if I try to use an empty tuple in a Result I get:

```
error[E0605]: non-primitive cast: `()` as `*mut c_void`
  --> ios/rust/src/lib.rs:81:1
   |
81 | #[swift_bridge::bridge]
   | ^^^^^^^^^^^^^^^^^^^^^^^ an `as` expression can only be used to convert between primitive types or to coerce to a specific trait object
   |
```

I'm working around it by using an empty `None` struct for now, but it'd be nice to convert this to a Swift function returning void.

chinedufn:
Thanks for the report.

I'd like to change this issue to be about the particular kind of `E` that you are using in your `Result<(), E>`. (i.e. opaque type, shared struct, shared enum, primitive, etc..).

The reason is that since `()` does not need an FFI representation (since the type always has the same value), we can add special case optimizations for some kinds of `Result<(), E>`.

For example, opaque types such as `type SomeType` get bridged as `*mut SomeType`.

This `*mut SomeType` pointer is never null, so `Result<(), SomeType>` can be bridged with a single pointer`*mut SomeType`, where `()` gets encoded as a null pointer.

I'd like to implement things that people are running into in real code, vs. tracking/implementing support for all `Result<(), ...>` in one issue.

Just giving you some context.

---

Anyways, what kind of `E` are you using? If it's an opaque type mind changing the issue name to `Support for Result<(), OpaqueType>`.

From there I can write up a thorough `good first issue` guide so that anyone that comes across this issue and is interested can quickly add support.

jfaust:
Well, I've been updating my bindings to use the new `Result` support. At the moment I'm using `Result<(), OpaqueType>` - however, I'd love to use an enum when that becomes possible (https://github.com/chinedufn/swift-bridge/issues/159).

I'll update this issue to be for `Result<(), OpaqueType>`. This is actually something I may try to tackle & submit a PR, as using an empty struct results in a lot of `let _ = ...` on the Swift side.

chinedufn:
Returning `Result<(), OpaqueType>` was implemented in https://github.com/chinedufn/swift-bridge/pull/180

I'll close this out. We can handle `Result<(), OpaqueType>` separately.

Thanks for opening and solving this!

qui-jason:
Awesome, we needed this today! Thanks @jfaust and @chinedufn!

PR #180, opened 2023-02-24
Support returning Result<(), OpaqueRust> from Rust functions

jfaust:
Adds partial support for #177 - supports returning `Result<(), OpaqueRust>` but not passing as an arg.

The change to `examples/rust-binary-calls-swift-package/build.rs` was to support xcode not existing in the default location (I use the `xcodes` tool).

This PR also supports empty types, based on @chinedufn's recommendation.

This enables the following:
```rust
#[swift_bridge::bridge]
mod ffi {
    struct UnitType;
    extern "Rust" {
        type OpaqueType;
        fn null_result() -> Result<(), OpaqueType>;
        fn unit_result() -> Result<UnitType, OpaqueType>;
    }
}

fn null_result() -> Result<(), OpaqueType> {
    Ok(())
}

fn unit_result() -> Result<UnitType, OpaqueType> {
    Ok(UnitType {})
}
```

### DIFF
--- a/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/ResultTests.swift
+++ b/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/ResultTests.swift
@@ -40,4 +40,16 @@ class ResultTests: XCTestCase {
             .Err(ResultTestOpaqueSwiftType(val: 666))
         )
     }
+
+    /// Verify that we can receive a Result<(), OpaqueRust> from Rust
+    func testSwiftCallRustResultNullOpaqueRust() throws {
+        try! rust_func_return_result_null_opaque_rust(true)
+
+        do {
+            try rust_func_return_result_null_opaque_rust(false)
+            XCTFail("The function should have returned an error.")
+        } catch let error as ResultTestOpaqueRustType {
+            XCTAssertEqual(error.val(), 222)
+        }
+    }
 }
--- a/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/ResultTests.swift
+++ b/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/ResultTests.swift
@@ -52,4 +52,16 @@ class ResultTests: XCTestCase {
             XCTAssertEqual(error.val(), 222)
         }
     }
+
+    /// Verify that we can receive a Result<UnitStruct, OpaqueRust> from Rust
+    func testSwiftCallRustResultUnitStructOpaqueRust() throws {
+        try! rust_func_return_result_unit_struct_opaque_rust(true)
+
+        do {
+            try rust_func_return_result_unit_struct_opaque_rust(false)
+            XCTFail("The function should have returned an error.")
+        } catch let error as ResultTestOpaqueRustType {
+            XCTAssertEqual(error.val(), 222)
+        }
+    }
 }
--- a/crates/swift-bridge-ir/src/bridged_type.rs
+++ b/crates/swift-bridge-ir/src/bridged_type.rs
@@ -58,7 +58,7 @@ pub(crate) trait BridgeableType: Debug {
         !self.is_built_in_type()
     }
 
-    /// Whether or not this type can be encoded to exactly one representation,;
+    /// Whether or not this type can be encoded to exactly one representation,
     /// and therefore can be encoded with zero bytes.
     /// For example `()` and `struct Foo;` can have exactly one representation,
     /// but `u8` cannot since there are 255 possible `u8`s.
--- a/crates/swift-bridge-ir/src/bridged_type.rs
+++ b/crates/swift-bridge-ir/src/bridged_type.rs
@@ -58,10 +58,11 @@ pub(crate) trait BridgeableType: Debug {
         !self.is_built_in_type()
     }
 
-    /// Whether or not this type can be encoded to exactly one representation.
+    /// Whether or not this type can be encoded to exactly one representation,;
+    /// and therefore can be encoded with zero bytes.
     /// For example `()` and `struct Foo;` can have exactly one representation,
     /// but `u8` cannot since there are 255 possible `u8`s.
-    fn has_exactly_one_encoding(&self) -> bool {
+    fn can_be_encoded_with_zero_bytes(&self) -> bool {
         self.only_encoding().is_some()
     }
 
--- a/crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
+++ b/crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
@@ -66,7 +66,7 @@ impl BuiltInResult {
                     Ok(ok) => {
                         #swift_bridge_path::result::ResultVoidAndPtr {
                             is_ok: true,
-                            err: 0 as *mut std::ffi::c_void
+                            err: std::ptr::null_mut::<std::ffi::c_void>()
                         }
                     }
                     Err(err) => {
--- a/crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
+++ b/crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
@@ -24,7 +24,7 @@ impl BuiltInResult {
         // TODO: Choose the kind of Result representation based on whether or not the ok and error
         //  types are primitives.
         //  See `swift-bridge/src/std_bridge/result`
-        let result_kind = if self.ok_ty.is_null() {
+        let result_kind = if self.ok_ty.can_be_encoded_with_zero_bytes() {
             quote! {
                 ResultVoidAndPtr
             }
@@ -60,7 +60,7 @@ impl BuiltInResult {
             span,
         );
 
-        if self.ok_ty.is_null() {
+        if self.ok_ty.can_be_encoded_with_zero_bytes() {
             quote! {
                 match #expression {
                     Ok(ok) => {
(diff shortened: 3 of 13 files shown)


crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
```diff
@@ -24,8 +24,14 @@ impl BuiltInResult {
         // TODO: Choose the kind of Result representation based on whether or not the ok and error
         //  types are primitives.
         //  See `swift-bridge/src/std_bridge/result`
-        let result_kind = quote! {
-            ResultPtrAndPtr
+        let result_kind = if self.ok_ty.is_null() {
```
chinedufn, line 27: Can use `self.ok_ty.has_exactly_one_encoding()` so that this works for types like `struct UnitStruct;`

I actually plan to rename `.has_exactly_one_encoding()` -> `.can_be_encoded_with_zero_bytes()`. 
Feel free to rename it while you're here if you like.

crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
```diff
@@ -129,19 +154,31 @@ impl BuiltInResult {
...
-            self.err_ty
-                .convert_ffi_expression_to_swift_type("val.ok_or_err!", type_pos, types);
-
-        format!(
-            "try {{ let val = {expression}; if val.is_ok {{ return {ok} }} else {{ throw {err} }} }}()",
-            expression = expression,
-            ok = convert_ok,
-            err = convert_err
-        )
+        if self.ok_ty.is_null() {
```
chinedufn, line 157:
```rust
if let Some(zero_byte_encoding) = self.ok_ty.only_encoding() {
    let ok = zero_byte_encoding.swift;
    // ... you can use `return {ok}` instead of `return`
}
```

Same idea for the other `.is_null()` calls.

src/std_bridge/result.rs
```diff
@@ -44,3 +44,12 @@ pub struct ResultPtrAndPtr {
     pub is_ok: bool,
     pub ok_or_err: *mut std::ffi::c_void,
 }
+
+// Bridges `Result<(), E>` where `E` is able to be encoded
```
chinedufn, line 48: Mind updating this comment to make it clear that this can also be used for types that don't require any bytes to encode such as `struct UnitStruct;`

crates/swift-bridge-ir/src/codegen/codegen_tests/result_codegen_tests.rs
```diff
@@ -263,3 +263,73 @@ void __swift_bridge__$some_function(struct __private__ResultPtrAndPtr arg);
         .test();
     }
 }
+
+/// Test code generation for Rust function that accepts a Result<T, E> where T is (), and E is an
```
chinedufn, line 267: Mind adding one more test module that tests `Result<struct UnitStruct;, OpaqueRust>`?

This way we can be confident that the suggestions above work.

SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/ResultTests.swift
```diff
@@ -40,4 +40,16 @@ class ResultTests: XCTestCase {
...
         )
     }
+
+    /// Verify that we can receive a Result<(), OpaqueRust> from Rust
+    func testSwiftCallRustResultNullOpaqueRust() throws {
+        try! rust_func_return_result_null_opaque_rust(true)
+
+        do {
+            try rust_func_return_result_null_opaque_rust(false)
+            XCTFail("The function should have returned an error.")
```
chinedufn, line 50: Nice!

chinedufn (approved):
This is AWESOME work, thanks.

Left some minor comments then we can merge this.

---

If you don't mind, how did you figure out where/how to implement this?

I'm asking because the goal is to make it easier for folks to contribute, so I'm curious about how hard or easy it was to make your first contribution?

Cheers and thanks!

examples/rust-binary-calls-swift-package/build.rs
```diff
@@ -24,8 +24,20 @@ fn main() {
     // ld: warning: Could not find or use auto-linked library 'swiftCompatibility50'
     // ld: warning: Could not find or use auto-linked library 'swiftCompatibilityDynamicReplacements'
     // ld: warning: Could not find or use auto-linked library 'swiftCompatibilityConcurrency'
-    println!("cargo:rustc-link-search={}",
-        "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/macosx/"
+    let xcode_path = if let Ok(output) = std::process::Command::new("xcode-select")
```
chinedufn, line 27: Awesome!

chinedufn:
@jfaust mind updating your PR body to include an example of what this PR makes possible?

Here are some other PRs to use as inspiration:

https://github.com/chinedufn/swift-bridge/pull/169
https://github.com/chinedufn/swift-bridge/pull/158
https://github.com/chinedufn/swift-bridge/pull/178

jfaust:
@chinedufn happy to make these changes, though it may take me a couple days to get to them.

Honestly making the changes was fairly straightforward - I found `result.rs`, read through it and got a bare-minimum understanding of what it was doing, and then grepped the code for `ResultPtrAndPtr`, which found most of the places I needed to make changes. From there it was mostly copy/paste with minor changes from the other `Result` code, which made it pretty easy. 

I was testing it all using my codebase, so once it was working I found the rust & swift `Result` tests, and added test cases to them.

I'm guessing this was probably one of the easier changes to make as a first-time contributor.

`swift-analyzer`'s "Expand macro recursively" command in vscode was also a huge help.

Thanks for building/maintaining this project!

jfaust:
@chinedufn I think I've resolved all your comments. Let me know if there's anything else to do!

crates/swift-bridge-ir/src/bridged_type.rs
```diff
@@ -58,10 +58,11 @@ pub(crate) trait BridgeableType: Debug {
         !self.is_built_in_type()
     }
 
-    /// Whether or not this type can be encoded to exactly one representation.
+    /// Whether or not this type can be encoded to exactly one representation,;
+    /// and therefore can be encoded with zero bytes.
```
chinedufn, line 62: nice

crates/swift-integration-tests/src/result.rs
```diff
@@ -68,6 +77,24 @@ fn rust_func_takes_result_opaque_swift(
...
+    } else {
+        Err(ResultTestOpaqueRustType { val: 222 })
+    }
+}
+
+fn rust_func_return_result_unit_struct_opaque_rust(
+    succeed: bool,
+) -> Result<ffi::UnitStruct, ResultTestOpaqueRustType> {
+    if succeed {
+        Ok(ffi::UnitStruct {})
```
chinedufn, line 92: Does this compile? Seems like it should be `ffi::UnitStruct` without the braces

crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
```diff
@@ -129,19 +154,32 @@ impl BuiltInResult {
...
-            self.err_ty
-                .convert_ffi_expression_to_swift_type("val.ok_or_err!", type_pos, types);
-
-        format!(
-            "try {{ let val = {expression}; if val.is_ok {{ return {ok} }} else {{ throw {err} }} }}()",
-            expression = expression,
-            ok = convert_ok,
-            err = convert_err
-        )
+        if let Some(zero_byte_encoding) = self.ok_ty.only_encoding() {
```
chinedufn, line 157: Minor, can we have a single `format!` instead of 2?

```rust
let (ok, err) = if let Some(zero_byte_encoding) = self.ok_ty.only_encoding() {
   // ...
} else {
  // ...
}

            format!(
                "try {{ let val = {expression}; if val.is_ok {{ return {ok} }} else {{ throw {err} }} }}()",
            )
```

chinedufn (approved):
Looks great!! Once tests are passing we can merge!

crates/swift-integration-tests/src/result.rs
jfaust: Huh - I didn't know you could do that. Changed, because why not, but yes, this does compile.

crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
jfaust: Done

jfaust:
@chinedufn I'm not sure why the Swift tests aren't compiling - they work fine on my machine with Xcode 14.2. Do you know what version of xcode is in use in the Action?

chinedufn:
I see this in the CI output

>  /Applications/Xcode_13.2.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/swift-frontend

So I'm assuming we're on 13.2.1 in CI.

I'm seeing this as the error:

> Unable to infer complex closure return type; add explicit type to disambiguate

Hmm...

chinedufn:
I'm on `Xcode 13.3.1` and can reproduce the error locally.

image

chinedufn:
Hmm.. so this fails to compile:

```swift
func foo() {
    { let foo = 5; return () }()
}
```

But this compiles successfully:

```swift
func foo() {
    { let foo = 5; return }()
}
```

chinedufn:
Ok, looks like it's possible to add an explicit signature at the beginning of the closure.

For example, this compiles:

```swift
struct MyStruct {}

func foo() {
    { () -> () in let foo = 5; return () }()
}

func bar() {
    { () -> MyStruct in let foo = 5; return MyStruct() }()
}
```

---

So our options (assuming we want to continue to support Xcode 13) are to either:

1. For the null type use `return` instead of `return ()`, along with a comment explaining that it's due to an Xcode 13 bug
2. For the null type use an explicit `() -> () in` at the beginning of the closure, along with a comment explaining that it's due to an Xcode 13 bug
3. OR, add the explicit `() -> ReturnType in` to the beginning of the closure regardless of the type

---

I would have assumed that having explicit types always leads to better compile times, but I did some quick googling (only looked at one result) and found that using explicit types was slower in the one link that I read (did not verify results myself, did not do thorough research, this post is two years old) https://forums.swift.org/t/regarding-swift-type-inference-compile-time-performance/49748/16

So... for now I like option 2.

Which boils down to:

```rust
let mut maybe_closure_type = "";

// There is a Swift compiler bug in Xcode 13 where using an explicit `()` here somehow leads
// the Swift compiler to a compile time error:
// "Unable to infer complex closure return type; add explicit type to disambiguate"
//
// It's asking us to add a `{ () -> () in .. }` explicit type to the beginning of our closure.
//
// So, for now, we remove the explicit `()` and instead 
if self.is_null() {
    maybe_closure_type = " () -> () in "
}

format!("try {{{maybe_closure_type} let val = {expression}; if val.is_ok {{ return {ok} }} else {{ throw {err} }} }}()")
```

On line 176 in `bridgeable_result.rs`

Issue #159, opened 2023-02-09
Support `Result<T, E>` where `E` is enum type.


NiwakaDev:
related to #152?

```rust
#[swift_bridge::bridge]
mod ffi {
    enum NetworkError {
        Authentication(String),
        Unknown,
        JsonParse,
    }

    extern "Rust" {
        type Data;
        async fn fetch_from_server() -> Result<Data, NetworkError>;
    }
}
```

NiwakaDev:
As far as I understand, I guess `swift-bridge` doesn't work right now.

chinedufn:
By the way, enums with data are not yet supported. Just created a new issue for that https://github.com/chinedufn/swift-bridge/issues/160

---

Right now, the `T` and `E` in a `Result<T, E>` only work for types that are passed between Swift and Rust as pointers.

We support this using https://github.com/chinedufn/swift-bridge/blob/420ed791c0cef08f7cfff7f4024ee39cd13d31c2/src/std_bridge/result.rs#L1-L7 and https://github.com/chinedufn/swift-bridge/blob/420ed791c0cef08f7cfff7f4024ee39cd13d31c2/crates/swift-bridge-build/src/generate_core/result_support.rs#L38-L40

Primitives, structs and enums aren't passed over FFI as pointers.

Primitives get passed as `C` primitives: 

https://github.com/chinedufn/swift-bridge/blob/07a15d78782dc371a7d24301ac9625124e42f8d8/crates/swift-bridge-ir/src/codegen/generate_c_header.rs#L453-L464

Structs and enums get passed as `C` structs that we generate at compile time:

https://github.com/chinedufn/swift-bridge/blob/c1fd7177b5d3a8918e34b6bc96ae29437cfc1487/crates/swift-bridge-ir/src/codegen/codegen_tests/transparent_struct_codegen_tests.rs#L78-L81

https://github.com/chinedufn/swift-bridge/blob/819e39160ec5b38dce2e60e2d39d078617ad158d/crates/swift-bridge-ir/src/codegen/codegen_tests/transparent_enum_codegen_tests.rs#L101-L105

---

So, to support `Result<T, E>` where either `T` or `E` isn't a pointer we need to generate an FFI representation on the fly like we currently do for transparent `struct`s and `enum`s.

---

Does that make sense?

NiwakaDev:
>By the way, enums with data are not yet supported. Just created a new issue for that https://github.com/chinedufn/swift-bridge/issues/160

[...]

PR #181, opened 2023-02-26
Support opaque types as enum variant fields 

NiwakaDev:
This PR adds support for opaque types as enum variant fields. However, This PR doesn't support generic opaque types. I'll write the reason later.

Here's an example:

```rust
#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        #[swift_bridge(Equatable)]
        type Foo;
        #[swift_bridge(associated_to = Foo)]
        fn new() -> Foo;
    }

    enum EnumWithNamedData {
        Variant1 { hello: String, data_u8: u8 },
        Variant2 { data_i32: i32 },
        Variant3 { foo: Foo },
    }

    extern "Rust" {
        fn reflect_enum_with_named_data(arg: EnumWithNamedData) -> EnumWithNamedData;
    }
}
```

```Swift
let enumWithNamedData3 = EnumWithNamedData.Variant3(foo: Foo.new())
switch reflect_enum_with_named_data(enumWithNamedData3) {
case .Variant3(let foo):
    XCTAssertEqual(foo, Foo.new())
    break
default:
    XCTFail()
}
```

### DIFF
--- a/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/SharedEnumTests.swift
+++ b/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/SharedEnumTests.swift
@@ -33,11 +33,11 @@ class SharedEnumTests: XCTestCase {
     }
 
     func testEnumWithUnnamedData() {
-        let enumWithUnnamedData1 = EnumWithUnnamedData.Variant1(create_string("hello"), Foo.new())
+        let enumWithUnnamedData1 = EnumWithUnnamedData.Variant1(create_string("hello"), Foo())
         switch reflect_enum_with_unnamed_data(enumWithUnnamedData1) {
         case .Variant1(let rustString, let foo):
             XCTAssertEqual(rustString.toString(), "hello")
-            XCTAssertEqual(foo, Foo.new())
+            XCTAssertEqual(foo, Foo())
         default:
             XCTFail()
         }
@@ -78,10 +78,10 @@ class SharedEnumTests: XCTestCase {
             XCTFail()
         }
 
-        let enumWithNamedData3 = EnumWithNamedData.Variant3(foo: Foo.new())
+        let enumWithNamedData3 = EnumWithNamedData.Variant3(foo: Foo())
         switch reflect_enum_with_named_data(enumWithNamedData3) {
         case .Variant3(let foo):
-            XCTAssertEqual(foo, Foo.new())
+            XCTAssertEqual(foo, Foo())
             break
         default:
             XCTFail()
--- a/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/SharedEnumTests.swift
+++ b/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/SharedEnumTests.swift
@@ -85,6 +85,23 @@ class SharedEnumTests: XCTestCase {
         default:
             XCTFail()
         }
-
+    }
+    
+    func testEnumWithOpaqueRust() {
+        let named = EnumWithOpaqueRust.Named(data: OpaqueRustForEnumTest())
+        switch reflect_enum_with_opaque_type(named) {
+        case .Named(let value):
+            XCTAssertEqual(value, OpaqueRustForEnumTest())
+        case .Unnamed(_):
+            XCTFail()
+        }
+        
+        let unnamed = EnumWithOpaqueRust.Unnamed(OpaqueRustForEnumTest())
+        switch reflect_enum_with_opaque_type(unnamed) {
+        case .Named(_):
+            XCTFail()
+        case .Unnamed(let value):
+            XCTAssertEqual(value, OpaqueRustForEnumTest())
+        }
     }
 }
--- a/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/SharedEnumTests.swift
+++ b/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/SharedEnumTests.swift
@@ -33,11 +33,11 @@ class SharedEnumTests: XCTestCase {
     }
 
     func testEnumWithUnnamedData() {
-        let enumWithUnnamedData1 = EnumWithUnnamedData.Variant1(create_string("hello"), 0)
+        let enumWithUnnamedData1 = EnumWithUnnamedData.Variant1(create_string("hello"), Foo.new())
         switch reflect_enum_with_unnamed_data(enumWithUnnamedData1) {
-        case .Variant1(let rustString, let valueUInt32):
+        case .Variant1(let rustString, let foo):
             XCTAssertEqual(rustString.toString(), "hello")
-            XCTAssertEqual(valueUInt32, 0)
+            XCTAssertEqual(foo, Foo.new())
         default:
             XCTFail()
         }
@@ -78,9 +78,10 @@ class SharedEnumTests: XCTestCase {
             XCTFail()
         }
 
-        let enumWithNamedData3 = EnumWithNamedData.Variant3
+        let enumWithNamedData3 = EnumWithNamedData.Variant3(foo: Foo.new())
         switch reflect_enum_with_named_data(enumWithNamedData3) {
-        case .Variant3:
+        case .Variant3(let foo):
+            XCTAssertEqual(foo, Foo.new())
             break
         default:
             XCTFail()
--- a/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/SharedEnumTests.swift
+++ b/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/SharedEnumTests.swift
@@ -33,27 +33,26 @@ class SharedEnumTests: XCTestCase {
     }
 
     func testEnumWithUnnamedData() {
-        let enumWithUnnamedData1 = EnumWithUnnamedData.Variant1(create_string("hello"), OpaqueRustForEnumTest())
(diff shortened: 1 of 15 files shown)


NiwakaDev:
When I am addressing #159, I noticed that `swift-bridge` doesn't support this feature yet.

NiwakaDev:
bridge module like:
```rust
enum SomeEnum {
    A(i32, SomeType<i32, u32>),
    B(String),
}
```

I guess that we want to generate the following same representation above (not ffi): 

```rust
#[derive ()]
pub enum SomeEnum {
    A(i32, super::SomeType<i32, u32>),
    B(String)
}
```

To generate `super::SomeType<i32, u32>`, we need to call the following method: 

https://github.com/chinedufn/swift-bridge/blob/b13c56ab399a9725bdb89b1435cafa76c056b21d/crates/swift-bridge-ir/src/bridged_type/bridged_opaque_type.rs#L37-L49

the above method doesn't currently generate `super::SomeType<i32, u32>`, and just generates `super::SomeType`. So, we need to pass `&TypeDeclarations` such as:

```rust
    fn to_rust_type_path(&self, types: &TypeDeclarations) -> TokenStream {
        let ty_name = &self.ty;
        let generics = self
        .generics
        .angle_bracketed_concrete_generics_tokens(types);

        if self.host_lang.is_rust() {
            quote! {
                super:: #ty_name #generics
            }
        } else {
            quote! {
                #ty_name
            }
        }
    }
```

chinedufn:
> I guess we can't to pass &TypeDeclarations as argument to to_tokens method.

Thanks for explaining.

We can delete the `impl ToTokens for PointerKind { .. }` and `impl ToTokens for Pointee { .. }` and instead create a `impl BridgeableType for OpaqueForeignType { .. }`

crates/swift-integration-tests/src/shared_types/shared_enum.rs
```diff
@@ -9,8 +9,15 @@ mod ffi {
         fn reflect_enum_with_no_data(arg: EnumWithNoData) -> EnumWithNoData;
     }
 
+    extern "Rust" {
+        #[swift_bridge(Equatable)]
+        type Foo;
+        #[swift_bridge(associated_to = Foo)]
```
chinedufn, line 15: Can use `#[swift_bridge(init)]`

crates/swift-integration-tests/src/shared_types/shared_enum.rs
```diff
@@ -9,8 +9,15 @@ mod ffi {
...
+        fn new() -> Foo;
+    }
+
     enum EnumWithUnnamedData {
-        Variant1(String, u32),
+        Variant1(String, Foo),
```
chinedufn, line 20: Let's use variant names that clearly indicate what we're testing.

Maybe something like:

```rust
enum EnumWithUnnamedData {
    TwoFields(String, u32),
    OneField(Foo),
    NoFields
}
```

crates/swift-bridge-ir/src/codegen/codegen_tests/transparent_enum_codegen_tests.rs
```diff
@@ -586,17 +592,20 @@ typedef struct __swift_bridge__$Option$SomeEnum { bool is_some; __swift_bridge__
...
         quote! {
             #[swift_bridge::bridge]
             mod ffi {
+                extern "Rust" {
+                    type SomeType;
+                }
                 enum SomeEnum {
                     A{
-                        data1: i32,
+                        data1: SomeType,
```
chinedufn, line 608: Instead of modifying these tests we can add a new test that is dedicated to opaque Rust data in enums

```rust
mod enum_with_opaque_rust_data {
// ...

#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        type SomeType;
    }

    enum SomeEnum {
        Unnamed(SomeType),
        Named { data: SomeType }
    }
}

// ...
}
```

---

Otherwise, just like how you just replace `i32` with `SomeType`, someone might come later and replace `SomeType` with `String`, not realizing that we put `SomeType` here intentionally.

SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/SharedEnumTests.swift
```diff
@@ -33,11 +33,11 @@ class SharedEnumTests: XCTestCase {
     }
 
     func testEnumWithUnnamedData() {
```
chinedufn, line 35: We can add a new `testEnumWithOpaqueRustData` that tests a named and unnamed variant

```rust
enum TestEnumWithOpaqueRust {
    Named { data: OpaqueRustForEnumTest },
    Unnamed(OpaqueRustForEnumTest)
}
```

crates/swift-integration-tests/src/shared_types/shared_enum.rs
```diff
@@ -22,7 +29,7 @@ mod ffi {
     enum EnumWithNamedData {
         Variant1 { hello: String, data_u8: u8 },
         Variant2 { data_i32: i32 },
-        Variant3,
+        Variant3 { foo: Foo },
```
chinedufn, line 32: Instead of adding in `Foo` to this, let's create a new enum each time we want to support a new kind of data.

```rust
enum TestEnumWithOpaqueRust {
    Named { data: OpaqueRustForEnumTest },
    Unnamed(OpaqueRustForEnumTest)
}
```

This way people can more easily see that we're explicitly trying to test opaque Rust data.

crates/swift-integration-tests/src/shared_types/shared_enum.rs
```diff
@@ -9,8 +9,15 @@ mod ffi {
         fn reflect_enum_with_no_data(arg: EnumWithNoData) -> EnumWithNoData;
     }
 
+    extern "Rust" {
+        #[swift_bridge(Equatable)]
+        type Foo;
```
chinedufn, line 14: Let's use a more descriptive name since if we use names like `Foo` the global namespace on the Swift side will get polluted with types that have names that aren't useful. 

Maybe `type OpaqueRustForEnumTest;`

chinedufn (changes requested):
Awesome work as usual.

Thanks for noticing this problem and then diving in to solve it. I really enjoy working with you.

---

I left some minor feedback. After that this looks good to me!

NiwakaDev:
Thank you for your quick reply!!

>We can delete the impl ToTokens for PointerKind { .. } and impl ToTokens for Pointee { .. } and instead create a impl BridgeableType for OpaqueForeignType { .. }

I guess `OpaqueForeignType` has implemented `BridgeableType` already. You mean "instead create a impl BridgeableType for BuiltInPointer { .. }"?

chinedufn:
Yup I meant `for BuiltInPointer`

NiwakaDev:
In my local environment, tests passed. I don't know why tests failed with github actions.

NiwakaDev:
In my local environment, tests passed. I don't know why tests failed with github actions.

In my local environment, I checked that the following tests passed.

```
test-swift-rust-integration.sh
```

```
test-swift-packages.sh
```

```
        RUSTFLAGS="-D warnings" cargo test -p swift-bridge \
          -p swift-bridge-build \
          -p swift-bridge-cli \
          -p swift-bridge-ir \
          -p swift-bridge-macro \
          -p swift-integration-tests
```

```
xcodebuild -project examples/codegen-visualizer/CodegenVisualizer/CodegenVisualizer.xcodeproj -scheme CodegenVisualizer
```

SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/SharedEnumTests.swift
```diff
@@ -33,59 +33,95 @@ class SharedEnumTests: XCTestCase {
...
+        let unnamed = EnumWithOpaqueRust.Unnamed(OpaqueRustForEnumTest())
+        switch reflect_enum_with_opaque_type(unnamed) {
+        case .Named(_):
+            XCTFail()
+        case .Unnamed(let value):
+            XCTAssertEqual(value, OpaqueRustForEnumTest())
+        }
+    }
 
+    func testEnumWithGenericOpaqueRust() {
```
chinedufn, line 107: Awesome

SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/SharedEnumTests.swift
```diff
@@ -33,59 +33,95 @@ class SharedEnumTests: XCTestCase {
...
+        let named = EnumWithGenericOpaqueRust.Named(data: new_generic_opaque_rust_for_enum_test())
+        switch reflect_enum_with_generic_opaque_type(named) {
+        case .Named(_):
+            //TODO: call several methods on GenericOpaqueRustForEnumTest<Int32>
```
chinedufn, line 111: Should include the reason that this is a TODO

```
// TODO: call a method on GenericOpaqueRustForEnumTest<Int32>
// after we add support for methods on generic opaque Rust Types.
// See https://github.com/chinedufn/swift-bridge/issues/44#issuecomment-1114198605
```

chinedufn (approved):
Looks great! I left a minor comment but it's fine if you don't do it. I'll merge once tests are passing.

chinedufn:
You're missing this in your `impl BridgeableType for BuiltInPointer`

https://github.com/chinedufn/swift-bridge/blob/master/crates/swift-bridge-ir/src/bridged_type.rs#L69-L72

> In my local environment, tests passed.

Maybe when you ran your tests your local branch is caught up with the latest master branch?

Tests shouldn't pass since the `only_encoding` method is required.

NiwakaDev:
Thank you!! Tests passed!!!

chinedufn:
Thanks!

PR #182, opened 2023-03-01
Support returning `Result<transparent enum, transparent enum>`  from Rust functions.

NiwakaDev:
This PR adds support for returning `Result<transparent enum, transparent enum>` from Rust functions.

related to #159.

```rust
#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        type ResultTestOpaqueRustType;
    }
    enum ResultTransparentEnum {
        Variant1,
        Variant2,
    }

    extern "Rust" {
        fn some_function() -> Result<ResultTestOpaqueRustType, ResultTransparentEnum>;
    }
}

```

```swift
do {
    let value = try some_function()
    //...
} catch {
    //...
}
```


NiwakaDev:
Note that this PR doesn't support returning from `async` function.

crates/swift-bridge-ir/src/bridged_type/bridgeable_custom_result.rs
```diff
@@ -0,0 +1,308 @@
+use crate::bridged_type::BridgedType;
+use crate::bridged_type::{BridgeableType, TypePosition, UnusedOptionNoneValue};
+use crate::{TypeDeclarations, SWIFT_BRIDGE_PREFIX};
+use proc_macro2::{Ident, Span, TokenStream};
+use quote::{format_ident, quote, ToTokens};
+use std::fmt::{Debug, Formatter};
+use syn::{Path, Type, TypeParam};
+
+#[derive(Clone)]
```
chinedufn, line 9: I don't think that we should have a separate type for custom results.

Fundamentally, we're bridging a Result.

Sure, whether or not we need to emit a custom struct is dependent on the inner `ok_ty` and `err_ty` combination, but that can be handled inside of the `BuiltInResult`.

So, instead of having a `BuiltInResult` and a `CustomResult` we should just have a `BuiltInResult`, then when the `ok_ty` and `err_ty` combination requires a custom result FFI representation  we can emit a `#[repr(C)]` support type.

---

I know that in https://github.com/chinedufn/swift-bridge/issues/159#issuecomment-1438383178 I suggested a custom result. I was wrong.

crates/swift-bridge-ir/src/codegen/codegen_tests/result_codegen_tests.rs
```diff
@@ -404,3 +404,175 @@ struct __private__ResultVoidAndPtr __swift_bridge__$some_function(void);
...
+                }
+            }
+        })
+    }
+
+    fn expected_swift_code() -> ExpectedSwiftCode {
+        ExpectedSwiftCode::ContainsAfterTrim(
+            r#"
+public func some_function() throws -> SomeOkType {
+    try {
```
chinedufn, line 456: The reason that the rest of the Swift codegen puts everything on one line is so that we can avoid needing to worry about indentation.

Technically indentation doesn't matter, but it feels wrong to me for us to emit code with "incorrect" indentation.

i.e. if you have a `Result<Result<T, E>, Foo>` the indentation would start to get weird since this uses a hard coded indentation of 4 spaces.

---

**So, I'd prefer if we just made this one line.**

And if we ever wanted to generate multiple lines we could introduce an indentation tracker so that the code that we emitted was nested nicely.

---

Let me know if you disagree.

crates/swift-bridge-ir/src/codegen/generate_rust_tokens.rs
```diff
@@ -250,6 +255,29 @@ impl ToTokens for SwiftBridgeModule {
                         }
                     };
                 }
+                TypeDeclaration::CustomResult(custom_result) => {
```
chinedufn, line 258: As mentioned above, I don't think that we need a separate `CustomResult` type. We can just use the `BuiltInResult`
Where the `BuiltIn` just means that we're using the Result from the stdlib. Then for certain ok/err combinations we emit some custom Rust and C code for the FFI representation.

This way any code that we write in the future that cares about results doesn't need to check both for built in results and custom results.

crates/swift-bridge-ir/src/parse.rs
```diff
@@ -117,6 +119,28 @@ impl Parse for SwiftBridgeModuleAndErrors {
                 if BridgedType::new_with_type(&unresolved_type, &type_declarations).is_some() {
                     continue;
                 }
+                if unresolved_type
```
chinedufn, line 122: This can go away if we just have a single `BuiltInResult`.

crates/swift-integration-tests/src/result.rs
```diff
@@ -36,6 +36,18 @@ mod ffi {
...
+    enum ResultTransparentEnum {
+        NamedField { data: i32 },
+        UnnamedFields(u8, String),
+        NoFields,
+    }
+
+    extern "Rust" {
+        fn rust_func_return_result_transparent_enum_opaque_rust(
+            succeed: bool,
+        ) -> Result<ResultTestOpaqueRustType, ResultTransparentEnum>;
```
chinedufn, line 49: Does `Result<Enum, Enum>` work?

If so, can we add a test for that?

Can we also test `Result<Enum, OpaqueRust>`, since that's a fairly common pattern in Rust code?

chinedufn (commented):
This is great!!!

My main feedback is around having a single `BuiltInResult` type instead of having `BuiltInResult + CustomResult`.

I know that I made the suggestion to use a `CustomResult` when we were originally discussing this. I was wrong. That was a bad idea. There's no reason to have 2 different types to store parsed results.
There is no need for the code generation concerns to bleed into the parsing code.

Other than that this looks awesome. Thanks!

crates/swift-integration-tests/src/result.rs
chinedufn: Can we also add a `fn same_custom_result_returned_twice_first() -> Result<Enum, Enum>` `fn same_custom_result_returned_twice_second() -> Result<Enum, Enum>` 

We don't need to call either of these from Swift. They're just here to confirm that we don't generate the same types twice.

crates/swift-bridge-ir/src/codegen/codegen_tests/result_codegen_tests.rs
```diff
@@ -404,3 +404,175 @@ struct __private__ResultVoidAndPtr __swift_bridge__$some_function(void);
...
+    fn extern_rust_result_transparent_enum_type_and_opaque_rust_type() {
+        CodegenTest {
+            bridge_module: bridge_module_tokens().into(),
+            expected_rust_tokens: expected_rust_tokens(),
+            expected_swift_code: expected_swift_code(),
+            expected_c_header: expected_c_header(),
+        }
+        .test();
+    }
+}
```
chinedufn, line 578: Can we add a `extern_rust_two_functions_return_same_custom_result` where we have two functions that both return `-> Result<OkEnum, ErrEnum>` and we confirm that the generated code only declares the custom enum once?

NiwakaDev:
>I think that during the codegen step when we're iterating over the functions we can keep track of the custom types that we encounter (i.e. in a HashSet or something) and then generate the supporting #[repr(C)] and C header code for them.

I see. Sounds good to me!! I think we can completely delete `CustomResultType` and `TypeDeclaration::CustomResult` with your idea.

>Let me know if you disagree with the idea of not having a separate CustomResult bridgeable type.

Actually, while I was implementing this feature, I thought about the following: 

*Maybe, we could generate custom result types without using TypeDeclarations, but we need to use it to generate custom result definitions. Hmm...*

Your idea could solve this problem. So, I agree with you!!

crates/swift-bridge-ir/src/bridged_type.rs
```diff
@@ -440,6 +444,17 @@ impl BridgeableType for BridgedType {
         }
     }
 
+    fn is_passed_via_pointer(&self) -> bool {
+        match self {
+            BridgedType::StdLib(ty) => match ty {
+                StdLibType::Null => true,
```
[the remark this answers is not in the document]
NiwakaDev: https://github.com/chinedufn/swift-bridge/blob/5acac1d4ba9269ed0de8d583a9c01509084ab197/crates/swift-bridge-ir/src/codegen/codegen_tests/result_codegen_tests.rs#L355-L377

In the above following, as far as I understand, `Null` seems to be passed over FFI via a pointer.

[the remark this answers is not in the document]
chinedufn: `Result<(), OpaqueType>` is bridged via pointer, but when `()` is bridged on its own it is not bridged via pointer.

Maybe we can rename this method to `is_always_bridged_via_pointer` to make this more clear?

---

`ResultVoidAndPtr` is used when you are bridging a type that is passed via pointer (such as `type MyRustType`) with a type that has a zero byte FFI encoding (such as `()` or `struct UnitStruct;`.

When you bridge a `()` on its own it isn't passed via pointer. Actually, it doesn't have any encoding on its own. Because it's a zero-sized type, we can "encode" it using zero bytes.

For a `Result<(), MyRustType>`, the `MyRustType` can be encoded as a non-null pointer, and we can use the null pointer to indicate that it is the `()` variant.

In fact `ResultVoidAndPtr` shouldn't exist at all since we can encode this into a pointer without the `is_ok` (a null pointer optimization). I just didn't catch it in code review https://github.com/chinedufn/swift-bridge/commit/27d7be991141586fb120e4691b7bea7d9eeecd17#r102860388 . (We don't have to fix it in this PR).

---

Here's how `()` is bridged when it's on its own and isn't inside of a `Result<T, E>`.

https://github.com/chinedufn/swift-bridge/blob/c15ba303a2c33dc457ddc2078cd0f8dbdc8e96de/crates/swift-bridge-ir/src/codegen/codegen_tests/single_representation_type_elision_codegen_tests.rs#L10-L82

NiwakaDev:
Addressed your review! Please let me know if there are any omissions or errors.

crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
```diff
@@ -220,16 +305,108 @@ impl BuiltInResult {
...
+            format!("struct __private__ResultPtrAndPtr")
         }
     }
+
+    pub fn generate_ffi_definition(
+        &self,
+        swift_bridge_path: &Path,
+        types: &TypeDeclarations,
+    ) -> Option<TokenStream> {
+        if self.is_custom_result_type() {
```
chinedufn, line 330: We can make the early return path more clear, as well as remove the indentation here by doing:

```rust
if !self.is_custom_result_type {
    return None;
}

// ...
```

crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
```diff
@@ -220,16 +305,108 @@ impl BuiltInResult {
...
+                    Ok(#ok),
+                    Err(#err),
+                }
+            });
+        }
+        return None;
+    }
+
+    pub fn generate_c_declaration(&self) -> Option<String> {
+        if self.is_custom_result_type() {
```
chinedufn, line 357: We can make the early return path more clear, as well as remove the indentation here by doing:

```rust
if !self.is_custom_result_type {
    return None;
}

// ...
```

crates/swift-bridge-ir/src/codegen/codegen_tests/result_codegen_tests.rs
```diff
@@ -404,3 +404,227 @@ struct __private__ResultVoidAndPtr __swift_bridge__$some_function(void);
...
+                }
+            }
+        }
+    }
+
+    fn expected_rust_tokens() -> ExpectedRustTokens {
+        ExpectedRustTokens::Contains(quote! {
+            #[repr(C)]
+            pub enum ResultVoidAndSomeErrEnum{
+                Ok(*mut std::ffi::c_void),
```
chinedufn, line 585: There's no reason for a `*mut std::ffi::c_void` here.

This can just be 

```rust
#[repr(C)]
pub enum ResultVoidAndSomeErrEnum {
    Ok,
    Err(__swift_bridge__SomeErrEnum)
}
```

crates/swift-bridge-ir/src/codegen/generate_c_header.rs
```diff
@@ -388,13 +395,17 @@ fn declare_func(
...
     types: &TypeDeclarations,
+    custom_type_declarations: &mut HashSet<String>,
 ) -> String {
     let ret = func.to_c_header_return(types);
     let name = func.link_name();
     let params = func.to_c_header_params(types);
 
     if let ReturnType::Type(_, ty) = &func.func.sig.output {
         if let Some(ty) = BridgedType::new_with_type(&ty, types) {
+            if let Some(declaration) = &ty.generate_c_declaration() {
```
chinedufn, line 406: Nice

crates/swift-bridge-ir/src/codegen/generate_rust_tokens.rs
```diff
@@ -275,12 +280,17 @@ impl ToTokens for SwiftBridgeModule {
                 }
             };
         }
-
+        let custom_type_definitions: Vec<TokenStream> = custom_type_definitions
```
chinedufn, line 283: This is fine, but `.into_values()` would also work here.

i.e.

```rust
let custom_type_definitions = custom_type_definitions.into_values();
let module_inner = quote! { 
    /* ... */ 

    #(#custom_type_definitions)*
}
```

should work.

---

Just a heads up in case you didn't know.

chinedufn (commented):
This is truly excellent work.

I left some feedback, but overall this is really fantastic.

I love it!!!

crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
NiwakaDev: How about making the early return in the others:

For example: 

https://github.com/NiwakaDev/swift-bridge/blob/8214e585cd222c2bfa03047997c191fcf55233e2/crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs#L205-L241

chinedufn: The one that you linked to has logic in both the true and false case, so it seems fine as is.

crates/swift-bridge-ir/src/bridged_type.rs
```diff
@@ -76,6 +77,19 @@ pub(crate) trait BridgeableType: Debug {
 
     fn as_result(&self) -> Option<&BuiltInResult>;
 
+    /// True if the type's FFI representation is a pointer
+    fn is_passed_via_pointer(&self) -> bool;
+
+    /// Generate the type's ffi definition if needed.
+    fn generate_ffi_definition(
```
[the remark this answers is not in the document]
NiwakaDev: >the idea being to use a word like "custom" to better reflect that this is meant for non built-in representations.

I see.  How about `generate_custom_rust_ffi_definition` and `generate_custom_c_ffi_declaration`?

NiwakaDev:
Addressed your review. But, I left one comment.

crates/swift-bridge-ir/src/bridged_type.rs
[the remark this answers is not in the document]
chinedufn: Sounds great

chinedufn:
Thanks!

PR #185, opened 2023-03-05
Support for returning `Result<OpaqueRustType, TransparentEnumType>` from `async` Rust functions

NiwakaDev:
This PR adds support for returning `Result<OpaqueRustType, TransparentEnumType>` from `async` Rust functions.

related to #159.

```rust
#[swift_bridge::bridge]
mod ffi {
    enum NetworkError {
	NoData,
	Authentication,
    }
    extern "Rust" {
        type Data;
    }
    extern "Rust" {
        async fn fetch_from_server() -> Result<Data, NetworkError>;
    }
}
```

```Swift
do {
    let data = try await fetch_from_server()
    //...
} catch let error as NetworkError {
    //...
} catch {
    //...
}
```

### DIFF
--- a/crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
+++ b/crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
@@ -410,23 +410,37 @@ typedef struct {c_enum_name}{{{c_tag_name} tag; union {c_fields_name} payload;}}
         return true;
     }
 
-    pub fn generate_async_run_wrapper_cb(&self, expression: &str, type_pos: TypePosition, types: &TypeDeclarations) -> String {
+    pub fn generate_async_run_wrapper_cb(
+        &self,
+        expression: &str,
+        type_pos: TypePosition,
+        types: &TypeDeclarations,
+    ) -> String {
         if self.is_custom_result_type() {
-            let ok = self.ok_ty.convert_ffi_expression_to_swift_type(&format!("{expression}.payload.ok"), type_pos, types);
-            let err = self.err_ty.convert_ffi_expression_to_swift_type(&format!("{expression}.payload.err"), type_pos, types);
+            let ok = self.ok_ty.convert_ffi_expression_to_swift_type(
+                &format!("{expression}.payload.ok"),
+                type_pos,
+                types,
+            );
+            let err = self.err_ty.convert_ffi_expression_to_swift_type(
+                &format!("{expression}.payload.err"),
+                type_pos,
+                types,
+            );
             return format!(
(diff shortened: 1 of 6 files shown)


crates/swift-integration-tests/src/async_function.rs
```diff
@@ -38,6 +38,30 @@ mod ffi {
         fn new(val: u32) -> AsyncResultOpaqueRustType2;
         fn val(&self) -> u32;
     }
+
+    enum AsyncResultOkEnum {
```
chinedufn, line 42: Do we need new async specific types here? 

I'm thinking that we can just re-use the existing opaque Rust types and enum Result types and we don't need `AsyncResult*` opaque types or enums.

crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
```diff
@@ -400,6 +409,44 @@ typedef struct {c_enum_name}{{{c_tag_name} tag; union {c_fields_name} payload;}}
 
         return true;
     }
+
+    pub fn generate_async_run_wrapper_cb(
```
chinedufn, line 413: `generate_swift_calls_async_rust_callback`

crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
```diff
@@ -228,7 +237,7 @@ impl BuiltInResult {
                     err_swift_type = err_swift_type
                 ),
                 TypePosition::SharedStructField => todo!(),
-                TypePosition::SwiftCallsRustAsyncOnCompleteReturnTy => todo!(),
+                TypePosition::SwiftCallsRustAsyncOnCompleteReturnTy => format!("なにわ"),
```
chinedufn, line 240: What's this?

chinedufn (commented):
Awesome!

Minor feedback then looks good to me.

crates/swift-integration-tests/src/async_function.rs
NiwakaDev: There were no transparent enums in async_function.rs. So, I appended new ones.

crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
NiwakaDev: Sorry... I forgot to remove this. I usually print some meaningless word to debug `swift-bridge`.

crates/swift-integration-tests/src/async_function.rs
chinedufn: Ohhhh ok

NiwakaDev:
Addressed your review. Please let me know if there are any omissions or errors

chinedufn:
Thanks!

Issue #168, opened 2023-02-13
Returning Result<struct, struct>


robinwit:
This seems to be supported, but I'm getting this error:
```
error[E0605]: non-primitive cast: `__swift_bridge__MyError` as `*mut c_void`
 --> src/lib.rs:6:1
  |
6 | #[swift_bridge::bridge]
  | ^^^^^^^^^^^^^^^^^^^^^^^ an `as` expression can only be used to convert between primitive types or to coerce to a specific trait object
  |
  = note: this error originates in the attribute macro `swift_bridge::bridge` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Source:
```
#![feature(extern_types)]

use crate::ffi::*;
use std::collections::HashMap;

#[swift_bridge::bridge]
mod ffi {
    #[swift_bridge(swift_repr = "struct")]
    pub struct MyIpAddress {
        pub origin: String,
    }

    #[swift_bridge(swift_repr = "struct")]
    pub struct MyError {}

    extern "Rust" {
        // type MyIpAddress;
        // type MyError;

        fn get_my_ip_from_rust() -> Result<MyIpAddress, MyError>;
    }
}

fn get_my_ip_from_rust() -> Result<MyIpAddress, MyError> {
    println!("Starting HTTP request from the Rust side...");

    let origin = reqwest::blocking::get("https://httpbin.org/ip")
        .unwrap()
        .json::<HashMap<String, String>>()
        .unwrap()
        .remove("origin")
        .unwrap();

    println!("HTTP request complete. Returning the value to Swift...");

    Ok(MyIpAddress { origin })
}
```

Am I missing something?

chinedufn:
`Result<T, E>` for structs isn't supported yet since we currently only support `Result` for types that are passed over FFI via pointers (such as opaque Rust types).

I think that this will be mostly solved by https://github.com/chinedufn/swift-bridge/issues/159 , since it will let us dynamically generate the FFI glue that we need for any supported `T` and `E`.

---

In short, this doesn't work yet, but should be easy to support after #159 lands. So feel free to subscribe to that issue.

I'll leave this issue open until we can confirm that it is solved.

\cc @NiwakaDev

robinwit:
Nice, thanks for the update! Wasn't sure that would fix it. :)

NiwakaDev:
We have added support returning `Result<enum, enum>` from Rust functions. 

So, We can implement returning `Result<struct, struct>` now.

[...]

PR #189, opened 2023-03-07
Support returning `Result<(), TransparentEnum>` from Rust async functions.

NiwakaDev:
This PR adds support for returning Result<(), TransparentEnum> from Rust async functions.

Here's an example of using this:

```Rust
//Rust
#[swift_bridge::bridge]
mod ffi {
    enum SomeEnum {
       //...
    }
    extern "Rust" {
        async fn some_function() -> Result<(), SomeEnum>;
    }
}
```

```Swift
//Swift
do {
    let value = try await some_function()
    //...
} catch let error as SomeEnum {
    //...
} catch {
    //...
}
```

### DIFF
--- a/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/AsyncTests.swift
+++ b/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/AsyncTests.swift
@@ -87,8 +87,6 @@ class AsyncTests: XCTestCase {
             XCTFail()
         } catch let error as AsyncResultErrEnum {
             switch error {
-            case .NoFields:
-                XCTFail()
             case .UnnamedFields(_, _):
                 XCTFail()
             case .NamedFields(let valueUInt32):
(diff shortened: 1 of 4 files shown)


NiwakaDev:
Note that This PR doesn't support returning `Result<Transparent, ()>` from Rust async functions. However, I guess that nobody would use returning `Result<Transparent, ()>` from Rust async functions.

NiwakaDev:
If this is merged, I guess that we can support #168.

chinedufn:
> However, I guess that nobody would use returning Result<TransparentEnum, ()> from Rust async functions.

Yeah I totally agree. 

It's unlikely that someone will run into this signature and if they do they can open an issue and use `#[swift_bridge(return_with = ...)]` to convert it into an `Option<TransparentEnum>` in the meantime.

> If this is merged, I guess that we can implement https://github.com/chinedufn/swift-bridge/issues/168.

Yeah!! They we can finally address this and make it a `Result<Option<TransparentStruct>, reqwest::Error>>` or something like that https://github.com/chinedufn/swift-bridge/blob/72c2fa5f2214e0845d9b3fa29b56ffbf94a97c68/examples/async-functions/src/lib.rs#L15-L17

chinedufn (approved):
Awesome. Very nice. Thanks for implementing this. 

Looks like your custom result implementation is holding up well since adding this support only took like 3 extra lines of code (excluding all of the tests),

Nice!

Issue #91, opened 2022-05-05
Vec of shared struct


smolck:
Currently something like the following Swift code:

```swift
class Test {
    func testFunction(things: RustVec<Thing>) { /* Do stuff */ }
}
```
and this Rust code:
```rust
#[swift_bridge::bridge]
mod ffi {
    #[swift_bridge(swift_repr = "struct")]
    pub struct Thing {
        pub some_str: String
    }

    extern "Swift" {
        type Test;

        #[swift_bridge(init)]
        fn new() -> Test;

        fn testFunction(&self, things: Vec<Thing>);
    }
}

// Then somewhere else
let test = ffi::Test::new();
test.testFunction(vec![Thing { some_str: "foo".to_string() }, /* ... */]);
```
Doesn't work, evidently due to the shared struct `Thing` not implmenting `Vectorizable` on the Swift side. Am I missing something and this sort of thing is possible? If not, could support for this be added?

chinedufn:
Hey.

Yup you're right that `Vec<SharedStruct>` isn't implemented yet. You aren't missing anything.

Right now we only support `Vec<OpaqueRustType>`

https://github.com/chinedufn/swift-bridge/blob/4d0a18f698c8244e874337f6fc9644b13f037e92/crates/swift-bridge-ir/src/codegen/codegen_tests/vec_codegen_tests.rs#L226-L236

https://github.com/chinedufn/swift-bridge/blob/b26dcf9c307d0bf77af96d5c29a6a479ad179a88/crates/swift-integration-tests/src/vec.rs#L1-L17

https://github.com/chinedufn/swift-bridge/blob/b26dcf9c307d0bf77af96d5c29a6a479ad179a88/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/VecTests.swift#L72-L81

---

Normally I'd try to get this to you right away but I'm pushing up against a deadline right now.

I can implement support for this over the weekend if that works for you?

Otherwise if it's urgent I'd be happy to provide guidance on how to implement this.

---

Thanks for opening this issue!

smolck:
> I can implement support for this over the weekend if that works for you?

Yes that’d be great! Thank you!

oeed:
I'm in need of this too, would be awesome to get added!

oeed:
@chinedufn this is blocking me currently so I'd be happy to try and implement, do you have any points as to how to do so?

chinedufn:
[...]

Issue #190, opened 2023-03-09
Using attributes on a transparent enum (specifically derive)


naturecodevoid:
Hi, this might be something I missed when reading the book/tests/examples but I can't find a way to use the `derive` attribute on a transparent enum. Here's an example:
```rs
#[swift_bridge::bridge]
mod ffi {
    #[derive(Debug)] // fails to compile unless this is removed
    pub enum TestEnum {
        Variant1,
        Variant2,
    }
}
```

Am I missing something or is this not supported? Are there any workarounds besides manually implementing the traits? Thanks.

naturecodevoid:
Do I have to manually implement the Debug part? Would I be able to easily use derive macros from other crates?

chinedufn:
We can consider supporting popular stdlib derives on a case by case basis (for example, we may want to support serde's derives).

But I don't think that we want to support arbitrary derives, since a derive could emit code that breaks out codegen.

This [comment in the cxx](https://github.com/dtolnay/cxx/issues/108#issue-598374322) explains a code example, where a derive macro could emit `type u32 = u8`. 
This would lead our codegen to emit code that it thinks is for a u32 when it's really for a u8, which could lead to undefined behavior if, for example, we have a `struct MyStruct { field: u32 }` which would be a `u8` on the Rust side but a `u32` on the Swift side.

---

> I will work on implementing this. Thanks for the detailed implementation guide!

Awesome!! Feel free to ask lots of questions as you dive in. I'm here to help!

---

> Also, if I'm correct, swift-bridge does not support copying docs from Rust to the generated Swift code. Is there an issue for this, and would I be able to help?

Yup, there is! It has an implementation guide that you could follow https://github.com/chinedufn/swift-bridge/issues/62

chinedufn:
Actually, it might be possible to safely support arbitrary derives. 
We'd need to think through all of the possible edge cases but in general I'm imagining something like:

---

Let's say we had this bridge module:

```rust
#[swift_bridge::bridge]
mod ffi {
    #[derive(SomeCustomDerive)]
    struct MyStruct {
        field: u32
    }
}
```

Which generated something like:

```rust
mod ffi {
    type u32 = ();

    struct MyStruct {
        field: u32
    }
}
```

We could make this a compile time error by emitting the following into every bridge module that used at least one deriving:

```rust
mod ffi {
    type u8 = swift_bridge::stdlib::u8;
    type i8 = swift_bridge::stdlib::i8;
    // ...
[...]

PR #193, opened 2023-03-11
Adding Vec support for transparent structs

rkreutz:
This PR addresses #91 

Added all the shims so transparent structs can support Vec. A couple caveats:
- In order to be able to implement the Vec methods in Rust, I needed to add a `#[derive(Clone)]` macro to the generated struct, which means that all properties in a struct must implement `Clone` or be able to derive it. We could maybe only add the macro if there are any Vec usage of the struct, but my knowledge on Rust is kind of limited to do so 😅 
- Vec support has only been implemented to `swift_repr = struct`, so `class`es still won't support Vec.

Great description on how to implement this in the issue @chinedufn, was very helpful.

### DIFF
--- a/crates/swift-bridge-ir/src/bridged_type/shared_struct.rs
+++ b/crates/swift-bridge-ir/src/bridged_type/shared_struct.rs
@@ -19,6 +19,13 @@ pub(crate) struct SharedStruct {
     pub fields: StructFields,
     pub swift_name: Option<LitStr>,
     pub already_declared: bool,
+    pub derives: StructDerives,
+}
+
+#[derive(Clone)]
+pub(crate) struct StructDerives {
+    pub copy: bool,
+    pub clone: bool,
 }
 
 impl SharedStruct {
--- a/crates/swift-bridge-ir/src/codegen/codegen_tests.rs
+++ b/crates/swift-bridge-ir/src/codegen/codegen_tests.rs
@@ -32,6 +32,7 @@ mod argument_label_codegen_tests;
 mod async_function_codegen_tests;
 mod boxed_fnonce_codegen_tests;
 mod conditional_compilation_codegen_tests;
+mod derive_struct_attribute_codegen_tests;
 mod extern_rust_function_opaque_rust_type_argument_codegen_tests;
 mod extern_rust_function_opaque_rust_type_return_codegen_tests;
 mod extern_rust_method_swift_class_placement_codegen_tests;
--- a/crates/swift-bridge-ir/src/codegen/codegen_tests/derive_struct_attribute_codegen_tests.rs
+++ b/crates/swift-bridge-ir/src/codegen/codegen_tests/derive_struct_attribute_codegen_tests.rs
@@ -1,0 +1,95 @@
+use super::{CodegenTest, ExpectedCHeader, ExpectedRustTokens, ExpectedSwiftCode};
+use proc_macro2::TokenStream;
+use quote::quote;
+
+/// Verify that we can derive the `Copy` trait on a transparent struct.
+mod derive_copy_struct {
+    use super::*;
+
+    fn bridge_module_tokens() -> TokenStream {
+        quote! {
+            #[swift_bridge::bridge]
+            mod ffi {
+                #[swift_bridge(swift_repr = "struct")]
+                #[derive(Copy)]
+                struct SomeStruct {
+                    field: u8,
+                }
+            }
+        }
+    }
+
+    fn expected_rust_tokens() -> ExpectedRustTokens {
+        ExpectedRustTokens::ContainsMany(vec![quote! {
+            #[derive(Copy)]
+            pub struct SomeStruct {
+                pub field: u8
+            }
+        }])
+    }
+
+    fn expected_swift_code() -> ExpectedSwiftCode {
+        ExpectedSwiftCode::SkipTest
+    }
+
+    fn expected_c_header() -> ExpectedCHeader {
+        ExpectedCHeader::SkipTest
+    }
+
+    #[test]
+    fn generates_struct() {
+        CodegenTest {
+            bridge_module: bridge_module_tokens().into(),
+            expected_rust_tokens: expected_rust_tokens(),
+            expected_swift_code: expected_swift_code(),
+            expected_c_header: expected_c_header(),
+        }
+        .test();
+    }
+}
+
+/// Verify that we can derive the `Clone` trait on a transparent struct.
+mod derive_clone_struct {
+    use super::*;
+
+    fn bridge_module_tokens() -> TokenStream {
+        quote! {
+            #[swift_bridge::bridge]
+            mod ffi {
+                #[swift_bridge(swift_repr = "struct")]
+                #[derive(Clone)]
+                struct SomeStruct {
+                    field: u8,
+                }
+            }
+        }
+    }
+
+    fn expected_rust_tokens() -> ExpectedRustTokens {
+        ExpectedRustTokens::ContainsMany(vec![quote! {
+            #[derive(Clone)]
+            pub struct SomeStruct {
+                pub field: u8
+            }
+        }])
+    }
+
+    fn expected_swift_code() -> ExpectedSwiftCode {
+        ExpectedSwiftCode::SkipTest
+    }
+
+    fn expected_c_header() -> ExpectedCHeader {
+        ExpectedCHeader::SkipTest
+    }
+
+    #[test]
+    fn generates_struct() {
+        CodegenTest {
+            bridge_module: bridge_module_tokens().into(),
+            expected_rust_tokens: expected_rust_tokens(),
+            expected_swift_code: expected_swift_code(),
+            expected_c_header: expected_c_header(),
+        }
+        .test();
+    }
+}
--- a/crates/swift-bridge-ir/src/codegen/generate_rust_tokens/shared_struct.rs
+++ b/crates/swift-bridge-ir/src/codegen/generate_rust_tokens/shared_struct.rs
@@ -90,7 +90,16 @@ impl SwiftBridgeModule {
             }
         };
 
+        let mut derives: Vec<TokenStream> = vec![];
+        if shared_struct.derives.copy {
+            derives.push(quote! {Copy});
+        }
+        if shared_struct.derives.clone {
+            derives.push(quote! {Clone});
+        }
+
         let definition = quote! {
+            #[derive(#(#derives),*)]
             pub struct #struct_name #struct_fields
 
             #struct_ffi_repr
--- a/crates/swift-bridge-ir/src/parse/parse_struct.rs
+++ b/crates/swift-bridge-ir/src/parse/parse_struct.rs
@@ -1,9 +1,12 @@
-use crate::bridged_type::{SharedStruct, StructFields, StructSwiftRepr};
+use crate::bridged_type::{
+    shared_struct::StructDerives, SharedStruct, StructFields, StructSwiftRepr,
+};
 use crate::errors::{ParseError, ParseErrors};
 use crate::parse::move_input_cursor_to_next_comma;
 use proc_macro2::Ident;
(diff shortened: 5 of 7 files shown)


chinedufn:
Thanks for putting this together!

I'll review this either tonight or tomorrow.

In the meantime, can you update your PR body to include an example of what this PR now enables?

These examples help with:
1. We use these examples in our release notes
2. We can link a new contributor to an old PR and they'll be able to more quickly understand what it's doing.

Here's an example of a PR that has an example in its body.

https://github.com/chinedufn/swift-bridge/pull/181

chinedufn:
> I needed to add a #[derive(Clone)] macro to the generated struct
> Vec support has only been implemented to swift_repr = struct, so classes still won't support Vec.

As you alluded to, requiring `Clone` on all structs wouldn't be great, so we should avoid that.

Right now transparent structs can't be passed by reference.

This is because neither I, nor any other contributor, has taken the time
to think through a good way to be able to pass Swift Structs to Rust by reference.

Swift classes, however, can easily be passed to Rust by reference. This is how sharing opaque types like `type SomeType` works.

I'm thinking that, for now, we only support `Vec<TransparentStruct>` when the struct has a class representation.

For example:

```rust
#[swift_bridge::bridge]
mod ffi {
    #[swift_bridge(swift_repr = "class")]
	struct Tomato {
        field: u32
	}
}
```

The above bridge module has `swift_repr = "class"`.

This would lead to us generating a Swift `class` instead of a Swift `struct`.
This should work in pretty much the exact same way as generating a class for an opaque Rust types.

```rust
#[swift_bridge::bridge]
mod ffi {
	extern "Rust" {
        // The codegen for the `struct Tomato` above should be pretty much the same as the
        // codegen for this opaque type.
        // In both cases we generate a Swift class
        // ( well, technically 3 -> https://chinedufn.github.io/swift-bridge/bridge-module/opaque-types/index.html#exposing-opaque-rust-types )
        type OpaqueTomato;
	}
}
```

---

`swift_repr = "class"` isn't supported yet. I think that we should add support for it (in a separate PR) and then make this PR support `swift_repr = "class"`
instead of `swift_repr = "struct"`.

The trick would be to massage the code so that when a struct has `swift_repr = "class"` it re-uses the same codegen code that we currently use for opaque Rust types.

Are you up for that? I'd be happy to write up a contribution guide if that would help.

rkreutz:
hey @chinedufn , so the issue behind me having to use `#[derive(Clone)]` was mostly due to Rust code, not really Swift to Rust bridge, here was the issue:
```rust
pub extern "C" fn _get(vec: *const Vec<#struct_name>, index: usize) -> #ffi_option_struct_repr {
    let vec = unsafe { &*vec };
    let val = vec.get(index).map(|v| v.clone()); // <---- I had to use .clone() here
    #ffi_option_struct_repr::from_rust_repr(val)
}
```
In contrast SharedEnum uses `.map(|v| *v)`, but I can't do the same on SharedStructs because the Rust compiler complaints about unborrowing a borrowed reference (something like that, not much familiar with Rust tbh). So if there were a way to pass this value along here in Rust I could remove the Clone derivation macro. Do you have any thoughts on this?

chinedufn:
The reason that it's working for enums is that right now we automatically emit `Copy, Clone` for non data-carrying enums. https://github.com/chinedufn/swift-bridge/blob/832498a8a822690e2ef61e5929a05d713ba2fbe8/crates/swift-bridge-ir/src/codegen/generate_rust_tokens/shared_enum.rs#L138

So, the `.map(|v| *v)` was using the enum's `Copy` implementation.

---

Copies are cheap so this is fine, but we don't want to have unnecessary `.clone()`s.

---

I suppose that we could land your implementation for transparent structs that explicitly implement Copy, but that would require us to add `#[derive(Copy, Clone)]` support to transparent structs.

Similar to https://github.com/chinedufn/swift-bridge/issues/190 https://github.com/chinedufn/swift-bridge/pull/194

---

I'd imagine that folks aren't using transparent structs that contain only `Copy` types all that often, so this is probably less useful than supporting `swift_repr = "class"`. 

Still, I would merge `swift_repr = "struct"` support if it was only implemented for `Copy` structs for now (until we eventually come up with a good way to pass `swift_repr = "struct"` structs by reference), and then we could worry about `swift_repr = "class"` structs later.

---

Does that all make sense, or no?

rkreutz:
It does, yes. One issue though, I believe it will be very common to have `String` properties in structs, and because of that they wouldn't be able to derive `Copy`. Should we allow deriving `Clone` in those cases and use it? We could try adding `derive(Copy)` if the struct supports it, and if not use `Clone`, would that be of any good?

Also, I think we will have this same issue for `swift_repr = "class"` right? Since in Rust it would still be a `struct` and this same snippet would fail, no?

rkreutz:
Also, it seems from the documentation that `Clone` is a supertrait of `Copy` and structs that would allow simple bitwise copies would use `*self` as the `.clone()` implementation, so on cases where the consumer is not using `String` in their structs this snippet should be pretty fast, no?

The main issue though is that starting from the release this PR would be merged, all structs exposed through the FFI would need to implement (or derive) `Clone`/`Copy`, which could be a breaking change for consumers of this library...

chinedufn:
Thank you for taking the time to think through all of this!

---

> Should we allow deriving Clone in those cases and use it?

I think we want to avoid needing to clone in order to pass things across the FFI boundary. 

Copy is fine, but I don't want us to encourage rampant cloning, especially if there's an alternative (`swift_repr = "class"`).

> The main issue though is that starting from the release this PR would be merged, all structs exposed through the FFI would need to implement (or derive) Clone/Copy,

Before landing this PR we would need to only generate the `Vec` impl for structs that `#[derive(Copy)]`.

So, this wouldn't be a breaking change since `Clone/Copy` wouldn't be a requirement (if we don't see a Copy derive we don't emit the vec support code.)

> Also, I think we will have this same issue for swift_repr = "class" right? Since in Rust it would still be a struct and this same snippet would fail, no?

No because we would pass a reference to the struct to Swift, and Swift would hold this reference inside of a class.

Here's a non swift-bridge example:

```rust
struct MyStruct {
    name: String
}

let instance = MyStruct { name: "hello".to_string() }'

let instance_ptr = &instance as *const MyStruct;

// We now have a pointer to the struct that we could pass over FFI.
```

This could is a good example of how passing a reference to a type from Rust -> Swift looks https://github.com/chinedufn/swift-bridge/blob/c1fd7177b5d3a8918e34b6bc96ae29437cfc1487/crates/swift-bridge-ir/src/codegen/codegen_tests/vec_codegen_tests.rs#L7-L171 .

Notice how we get a reference to the type, cast it to a pointer and return that pointer:

https://github.com/chinedufn/swift-bridge/blob/c1fd7177b5d3a8918e34b6bc96ae29437cfc1487/crates/swift-bridge-ir/src/codegen/codegen_tests/vec_codegen_tests.rs#L44-L53

---

I'd say that the best solution, at least for now, would be to support vecs on `swift_repr = "class"` structs.

But, I'd be cool with landing you `swift_repr = "struct"` support if it only worked on structs that had an explicit `derive(Copy, Clone)` (otherwise we don't emit the vec support code for the struct). 
Although I think we both agree that requiring `Copy` in order to bridge a struct doesn't sound very useful.

I don't think that we can land anything that relies on cloning though, since `swift-bridge` aims to be a high performance bridging library with either zero or as close to zero overhead as possible in all bridging scenarios.

rkreutz:
> Yeah, now that I think more about it figuring out how to best support `swift_repr = "class"` will require some design work.
> 
> So it might be best to land this PR with support for `swift_repr = "struct"` structs that derive Copy and Clone. (see this comment about `#[derive(Copy, Clone)]` [#193 (comment)](https://github.com/chinedufn/swift-bridge/pull/193#issuecomment-1467165955)

Yeah, I was thinking about it as well, the main issue about classes and pointers is how do you make sure that the pointer in Rust is aligned with Swift, I don't see any way of checking that, more so, how do you align them if they aren't. It seems shared classes aren't supported yet as well, so probably there is no answer to this yet.

So yeah, checking if the shared struct has `Copy` trait might be the best approach. However this will leave people with `String`s in their structs out of this change, since they can't implement `Copy` on them, which might be heavily used, but I'll focus on the task at hand now.

So how can I check if a shared struct implements the `Copy` trait? I realise I'd do the check on `/swift-bridge/crates/swift-bridge-ir/src/codegen/generate_rust_tokens/shared_struct.rs` but I have no idea what property I can check on `SharedStruct`

chinedufn:
> Yeah, I was thinking about it as well, the main issue about classes and pointers is how do you make sure that the pointer in Rust is aligned with Swift, I don't see any way of checking that, more so, how do you align them if they aren't. It seems shared classes aren't supported yet as well, so probably there is no answer to this yet.

I'm not understanding what you mean? This would essentially be an opaque pointer.

But, at any rate, I thought about this a little more and realized that `swift_repr = "class"` needs some dedicated design thinking. I've made the mistake in this thread of sharing some half baked ideas.. really instead I need to sit down and more deeply explore the design space and experiment a bit.

I'll make time for that at some point.

> So how can I check if a shared struct implements the Copy trait?

We'd have to add support for `#[derive(Copy, Clone)]` on structs.

Similar to this issue about enums, but for structs instead https://github.com/chinedufn/swift-bridge/issues/190 https://github.com/chinedufn/swift-bridge/pull/194

If you're up for it, you could land that as a separate PR and then you'd be able to check for `copy` in this PR.

rkreutz:
> If you're up for it, you could land that as a separate PR

@chinedufn opened the PR for `derive()` here https://github.com/chinedufn/swift-bridge/pull/198

> and then you'd be able to check for copy in this PR.

Ideally we should check whether the trait is implemented as well, since one can implement these traits manually without the `derive` macro. I'll leave a TODO in the code for such cases, but I guess for now checking the derives should be good enough.

chinedufn:
I suppose one could `impl Copy for MyStruct { }`, but since it's a marker trait I'm not sure if there is any value in doing it outside of the bridge module.

But yeah, it's certainly possible. If we ever wanted to handle this we could use a `#[swift_bridge(Copy)]` attribute to indicate that the struct is Copy.

```rust
#[swift_bridge::bridge]
mod ffi {
    #[swift_bridge(Copy)]
    struct MyStruct {}
}

impl Copy for ffi::MyStruct {}
```

We do something similar for opaque Rust types https://chinedufn.github.io/swift-bridge/bridge-module/opaque-types/index.html#swift_bridgecopysize

https://github.com/chinedufn/swift-bridge/blob/e7aef343da039d0ff9e2783cf7989443ed39cca3/crates/swift-bridge-ir/src/codegen/codegen_tests/generic_opaque_rust_type_codegen_tests.rs#L259-L284

> I'll leave a TODO in the code for such cases

Sounds good!

crates/swift-bridge-ir/src/codegen/codegen_tests/vec_codegen_tests.rs
```diff
@@ -824,6 +824,210 @@ void* __swift_bridge__$Vec_SomeStruct$as_ptr(void* vec_ptr);
     }
 }
 
+mod transparent_struct_vec_support_without_derives {
```
chinedufn, line 827: Minor request:

Mind adding `/// Verify that ...` doc comments to any codegen test modules that you add.

They can sometimes feel redundant but I find that they really help when trying to understand a test since it's easier to read and make sense of prose than long snake case identifiers.
Especially for new contributors that are trying to understand what's going on.

rkreutz: Done [b08764e](https://github.com/chinedufn/swift-bridge/pull/193/commits/b08764eb93e9627ea61db3036e2f6e4a672e2072)

rkreutz:
hey @chinedufn I ended up opening a separate PR because of some conflicts between my fork's `master`, you can have a look here https://github.com/chinedufn/swift-bridge/pull/199

I've added more details on the PR's description

PR #194, opened 2023-03-11
derive(Debug) enum implementation

naturecodevoid:
Related issue: #190 

Error handling is nonexistent and I had to remove/change some things to get it to compile with no warnings (such as adding `#[allow(dead_code)]` in `parse_error.rs` since `EnumUnrecognizedAttribute` isn't used) so the PR will need some small changes before merge

I was unsure on how to reorganize `enum_attributes.rs` so you might want me to rename some things. Additionally, in `SharedEnum`, I think it would be better to keep an `attributes: SharedEnumAllAttributes` instead of redefining all of those properties. What do you think?

Example:

```rs
#[swift_bridge::bridge]
mod ffi {
    #[derive(Debug)]
    enum DeriveDebugEnum {
        Variant,
    }
}
```

```swift
let debugString = String(reflecting: DeriveDebugEnum.Variant)
XCTAssertEqual(debugString, "Variant")
```


crates/swift-bridge-ir/src/codegen/generate_c_header.rs
```diff
@@ -148,6 +148,12 @@ typedef struct {option_ffi_name} {{ bool is_some; {ffi_name} val; }} {option_ffi
                             variants += &variant;
                         }
 
+                        let derive_debug_impl = if ty_enum.derive.debug {
+                            format!("void* {ffi_name}$_Debug({ffi_name} this);")
```
chinedufn, line 152: Should end the string with a newline `\n`.

crates/swift-bridge-ir/src/codegen/generate_rust_tokens/shared_enum.rs
```diff
@@ -130,14 +130,31 @@ impl SwiftBridgeModule {
...
-        let automatic_derives = if shared_enum.has_one_or_more_variants_with_data() {
+        // Auto derives
+        let mut derives = if shared_enum.has_one_or_more_variants_with_data() {
             vec![]
         } else {
             vec![quote! {Copy}, quote! {Clone}]
         };
 
+        // User derives
+        let mut derive_impls = vec![];
```
chinedufn, line 141: Let's change the variable name since reads as if it is storing `impl DerivedTrait for for MyEnum { .. }`.

Maybe call it `derive_impl_ffi_bridges`? If you have a better name in mind, go for it.

crates/swift-bridge-ir/src/errors/parse_error.rs
```diff
@@ -5,6 +5,7 @@ use syn::{ForeignItemFn, ForeignItemType, LitStr};
 use syn::{Token, Type};
 
 // <!-- ANCHOR: mdbook-parse-error-enum -->
+#[allow(dead_code)] // TODO: remove
```
chinedufn, line 8: Let's remove this

crates/swift-bridge-ir/src/parse/parse_enum/enum_attributes.rs
```diff
@@ -0,0 +1,104 @@
...
+
+        let attr = match key.to_string().as_str() {
+            "already_declared" => EnumAttr::AlreadyDeclared,
+            "swift_name" => {
+                input.parse::<Token![=]>()?;
+
+                let name = input.parse()?;
+                EnumAttr::SwiftName(name)
+            }
+            _ => {
```
chinedufn, line 96: Let's add a UI test for this:

We can first combine the existing unrecognized attribute tests into a single `unrecognized-attribute.rs` and `unrecognized-attribute.stderr` UI test instead of having 3 different files.

https://github.com/chinedufn/swift-bridge/blob/3c0d00da33fdf19e201eb121c7e6e370897f8070/crates/swift-bridge-macro/tests/ui/unrecognized-argument-attribute.rs

https://github.com/chinedufn/swift-bridge/blob/3c0d00da33fdf19e201eb121c7e6e370897f8070/crates/swift-bridge-macro/tests/ui/unrecognized-function-attribute.rs

https://github.com/chinedufn/swift-bridge/blob/3c0d00da33fdf19e201eb121c7e6e370897f8070/crates/swift-bridge-macro/tests/ui/unrecognized-opaque-type-attribute.rs

---

Then we can add an enum to the unrecognized attribute UI test file

---

Then we can get the test passing using something like:

```rust
#[derive(Default)]
pub(super) struct SharedEnumSwiftBridgeAttributes {
    pub already_declared: bool,
    pub swift_name: Option<LitStr>,
    pub errors: Vec<ParseError>
}

impl SharedEnumSwiftBridgeAttributes {
    pub(super) fn store_attrib(&mut self, attrib: EnumAttr) -> syn::Result<()> {
        match attrib {
            EnumAttr::AlreadyDeclared => self.already_declared = true,
            EnumAttr::Error(err) => { self.errors.push(error) },
            EnumAttr::SwiftName(name) => self.swift_name = Some(name),
        };
        Ok(())
    }
}
```

Then in https://github.com/chinedufn/swift-bridge/blob/b23f7d3416b740d0bd920c6edce8e42d8dfe3374/crates/swift-bridge-ir/src/parse/parse_enum.rs#L16-L42

We can do something like:

```rust
for err in attribs.errors {
    self.errors.push(err);
}
```

crates/swift-bridge-ir/src/codegen/generate_rust_tokens/shared_enum.rs
```diff
@@ -130,14 +130,31 @@ impl SwiftBridgeModule {
...
+
+        if shared_enum.derive.debug {
```
chinedufn, line 143: Can we, for now, add a `&& !shared_enum.has_one_or_more_variants_with_data()`.

Reason being that our current `Debug` implementation might lead people to making memory safety mistakes when using enums that hold non `Copy` data .

For example, say you have the following enum.

```rust
#[swift_bridge::bridge]
mod ffi {
    enum MyEnum {
        Variant(String)
    }
}
```

Right now we don't have a way to pass enums by reference. 
So, right now passing an enum from `Swift` -> `Rust` means passing ownership from `Swift` -> `Rust`.

So, if you were to call the enum's Debug implementation you would be passing ownership of the enum, and thus the enum's inner data, to Rust.

So, if you tried to use the inner data after calling the `Debug` implementation you would trigger undefined behavior.

---

At some point we'll want to design a way to pass enums between `Rust`/`Swift` by reference.. 
But for the purpose of this PR we can just limit `Debug` to enums that implement `Copy/Clone`.

crates/swift-bridge-ir/src/codegen/generate_rust_tokens/shared_enum.rs
```diff
@@ -130,14 +130,31 @@ impl SwiftBridgeModule {
...
         }
 
-        // TODO:
-        //  Parse any derives that the user has specified and combine those with our auto derives.
-        let automatic_derives = if shared_enum.has_one_or_more_variants_with_data() {
+        // Auto derives
+        let mut derives = if shared_enum.has_one_or_more_variants_with_data() {
             vec![]
         } else {
             vec![quote! {Copy}, quote! {Clone}]
```
chinedufn, line 137: Hmm... I wish we didn't automatically derive anything and instead did it explicitly.. Especially given by comment below about restricting this to `Copy` enums for now.. 

We can worry about this later though.. Just sharing a thought. Does not need to impact this PR.

chinedufn (commented):
Looks great!! Thanks for putting this together

Left some minor comments but overall this looks good to me.

Can you update your PR body to include an example of what this PR now enables?

These examples help with:
1. We use these examples in our release notes
2. We can link a new contributor to an old PR and they'll be able to more quickly understand what it's doing.

Here's an example of a PR that has an example in its body.

https://github.com/chinedufn/swift-bridge/pull/181

---

Thanks for putting this together!

crates/swift-bridge-ir/src/codegen/generate_c_header.rs
naturecodevoid: It generates just fine:
```h
// File automatically generated by swift-bridge.
#include <stdbool.h>
typedef enum __swift_bridge__$DeriveDebugEnumTag { __swift_bridge__$DeriveDebugEnum$Variant, } __swift_bridge__$DeriveDebugEnumTag;
typedef struct __swift_bridge__$DeriveDebugEnum { __swift_bridge__$DeriveDebugEnumTag tag; } __swift_bridge__$DeriveDebugEnum;
typedef struct __swift_bridge__$Option$DeriveDebugEnum { bool is_some; __swift_bridge__$DeriveDebugEnum val; } __swift_bridge__$Option$DeriveDebugEnum;
void* __swift_bridge__$DeriveDebugEnum$_Debug(__swift_bridge__$DeriveDebugEnum this);
void* __swift_bridge__$Vec_DeriveDebugEnum$new(void);
void __swift_bridge__$Vec_DeriveDebugEnum$drop(void* vec_ptr);
void __swift_bridge__$Vec_DeriveDebugEnum$push(void* vec_ptr, __swift_bridge__$DeriveDebugEnum item);
__swift_bridge__$Option$DeriveDebugEnum __swift_bridge__$Vec_DeriveDebugEnum$pop(void* vec_ptr);
__swift_bridge__$Option$DeriveDebugEnum __swift_bridge__$Vec_DeriveDebugEnum$get(void* vec_ptr, uintptr_t index);
__swift_bridge__$Option$DeriveDebugEnum __swift_bridge__$Vec_DeriveDebugEnum$get_mut(void* vec_ptr, uintptr_t index);
uintptr_t __swift_bridge__$Vec_DeriveDebugEnum$len(void* vec_ptr);
void* __swift_bridge__$Vec_DeriveDebugEnum$as_ptr(void* vec_ptr);
```

crates/swift-bridge-ir/src/errors/parse_error.rs
naturecodevoid: I added it there because I didn't think it was a good idea to remove the enum attribute error, and currently since error handling isn't implemented it's unused and therefore generates a warning.

crates/swift-bridge-ir/src/codegen/generate_rust_tokens/shared_enum.rs
naturecodevoid: [`e1a701a` (#194)](https://github.com/chinedufn/swift-bridge/pull/194/commits/e1a701a4973faef13a593d26d8270954edfbf598)

crates/swift-bridge-ir/src/codegen/generate_rust_tokens/shared_enum.rs
naturecodevoid: [`5aba738` (#194)](https://github.com/chinedufn/swift-bridge/pull/194/commits/5aba738af46992d65943ad6ec77d564069decffe)

naturecodevoid:
Sorry for taking a while to get back to you. I was adding the UI test, but I ran into an issue. With this UI test:
```rs
#[swift_bridge::bridge]
mod ffi {
    #[swift_bridge(InvalidAttribute)]
    enum SomeEnum {}

    extern "Rust" {
        #[swift_bridge(InvalidAttribute)]
        fn some_function();

        #[swift_bridge(InvalidAttribute)]
        type SomeType;

        fn some_function(#[swift_bridge(InvalidArgumentAttribute)] some_value: isize);
    }
}

pub struct SomeType;
```
this is the only error that would come up.
```
error: Unrecognized attribute "InvalidAttribute".
  --> tests/ui/unrecognized-attribute.rs:14:24
   |
14 |         #[swift_bridge(InvalidAttribute)]
   |                        ^^^^^^^^^^^^^^^^
```
So it seems to be short circuiting somehow. (it only reported InvalidAttribute for `type SomeType;`) If I only have the enum, the error is reported correctly.

chinedufn:
> So it seems to be short circuiting somehow

Gotcha

You can keep the other files as is and add an `unrecognized-enum-attribute.rs` test for now, and just leave a comment inside the test that we should eventually combine all of the unrecognized attribute tests into one file.

crates/swift-bridge-ir/src/codegen/generate_rust_tokens/shared_enum.rs
```diff
@@ -140,7 +140,8 @@ impl SwiftBridgeModule {
         // User derives
         let mut derive_impl_ffi_bridges = vec![];
 
-        if shared_enum.derive.debug {
+        // Current derive(Debug) implementation could cause memory safety issues: https://github.com/chinedufn/swift-bridge/pull/194#discussion_r1134386788
```
chinedufn, line 143: Minor tweak to make it clear that we're preventing the memory safety issue.

```rust
// We currently only allow derive(Debug) on non data carrying enums since
// to prevent a potential memory safety issue.
// https://github.com/chinedufn/swift-bridge/pull/194#discussion_r1134386788
```

chinedufn:
Left one tiny piece of feedback around the derive Debug comment then we can merge this (once the invalid attribute test passes).

Thanks for putting this all together!

naturecodevoid:
I've added the UI test, should be ready for merge unless there's anything else I need to change!

crates/swift-bridge-ir/src/parse/parse_enum/enum_attributes.rs
```diff
@@ -0,0 +1,125 @@
...
+                            _ => todo!("Unsupported derive macro; please see https://github.com/chinedufn/swift-bridge/issues/190#issuecomment-1463234027"),
+                        }
+                    }
+                }
+                "swift_bridge" => {
+                    attributes.swift_bridge = attr.parse_args()?;
+                    // note: this will empty attributes.swift_bridge.errors, maybe it would be better to clone?
+                    attributes
+                        .errors
+                        .append(&mut attributes.swift_bridge.errors);
```
chinedufn, line 40: Can delete the comment and leave it as is for now. 

I say this because if we commented on every possible trivial improvement the codebase might have hundreds of these. 
I think it's fine to expect a future maintainer to know how to refactor this if they want to make things more correct.

Nevertheless I appreciate your drive for correctness!

---

Alternatively, you could make an `AttrParseResult` enum type and do something like:

```rust
match attr.parse_args::<AttrParseResult<SharedEnumSwiftBridgeAttributes>>()? {
    AttrParseResult::Ok(ok) => attributes.swift_bridge = ok,
    AttrParseResult::Err(errors) => attributes.errors.append(&mut errors)
}
```

This way it becomes clear that there are either attributes, or errors, and there isn't the chance to check for `.errors` and see nothing since we drained them.

---

I'm fine with either approach since this is all fairly self-contained internal stuff.

chinedufn (approved):
LGTM once tests are passing.

You can run tests locally using `cargo test -p swift-bridge-ir`.

Looks like the failure is just an off by one line number error, so it's possible that you actually did run the tests locally but are just running a different version of the compiler than CI is.

I don't know, but either way just tweaking the .stderr file should fix it.

Failing here https://github.com/chinedufn/swift-bridge/actions/runs/4598004486/jobs/8125224915?pr=194

naturecodevoid:
Tests not passing was due to me adding the TODO comment after running the tests, therefore offsetting the error by 1 line. Should be fixed now 👍

chinedufn:
Thanks a lot for your contribution!

PR #198, opened 2023-03-16
Adding derive() macro support to SharedStructs

rkreutz:
This PR adds support for `#[derive(...)]` macros to `SharedStruct`s.

## Example

```rust
#[swift_bridge::bridge]
mod ffi {
    #[derive(Copy, Clone)]
    struct Foo {
        field: u8,
    }
}
```

This will successfully add a derived `Copy` and `Clone` trait to the struct declared.


crates/swift-bridge-ir/src/codegen/generate_rust_tokens/shared_struct.rs
```diff
@@ -90,7 +90,15 @@ impl SwiftBridgeModule {
             }
         };
 
+        let automatic_derives;
```
chinedufn, line 93: We're moving away from automatic derives, so `derives` is fine.

crates/swift-bridge-ir/src/bridged_type/shared_struct.rs
```diff
@@ -19,6 +19,7 @@ pub(crate) struct SharedStruct {
     pub fields: StructFields,
     pub swift_name: Option<LitStr>,
     pub already_declared: bool,
+    pub derives: Option<Vec<TokenStream>>,
```
chinedufn, line 22: Instead of `Option<Vec<TokenStream>>` let's use something like:

```rust
pub(crate) struct StructDerives {
    pub copy: bool,
    pub clone: bool,
}
```

crates/swift-bridge-ir/src/parse/parse_struct.rs
```diff
@@ -302,13 +320,33 @@ mod tests {
...
+    }
+
     /// Verify that we properly parse multiple comma separated struct attributes.
     #[test]
     fn parses_multiple_struct_attributes() {
         let tokens = quote! {
             #[swift_bridge::bridge]
             mod ffi {
                 #[swift_bridge(swift_name = "FfiFoo", swift_repr = "class")]
+                #[derive(Copy, Clone)]
```
chinedufn, line 349: Can we remove this derive?

crates/swift-bridge-ir/src/parse/parse_struct.rs
```diff
@@ -302,13 +320,33 @@ mod tests {
         assert_eq!(ty.swift_name.as_ref().unwrap().value(), "FfiFoo");
     }
 
+    /// Verify that we parse the derive(...)
+    #[test]
+    fn parse_derive_attribute() {
+        let tokens = quote! {
+            #[swift_bridge::bridge]
+            mod ffi {
+                #[derive(Copy, Clone)]
```
chinedufn, line 329: Nice

chinedufn (commented):
Nice looks great. Left a couple minor comments then we can land.

chinedufn:
Just missing a codegen test and an integration test (don't need the Swift side .. just a Rust integration test is fine for now and in the future we can emit a clone function on the Swift side...)

See the following files in https://github.com/chinedufn/swift-bridge/pull/194/files for an example that you can copy/paste/modify

- `crates/swift-bridge-ir/src/codegen/codegen_tests/derive_attribute_codegen_tests.rs`
- `crates/swift-integration-tests/src/enum_attributes/derive.rs`

crates/swift-bridge-ir/src/bridged_type/shared_struct.rs
rkreutz: This would kind of limit us to only these kind of derives, if we want to add new ones (like `Debug`) we would need to open a PR and add them. It also has the potential of growing quite a lot. I don't see why it would be an issue to have the more generic `Vec<TokenStream>`?

crates/swift-bridge-ir/src/parse/parse_struct.rs
rkreutz: I'm actually testing here whether `#[swift_bridge]` and `#[derive]` work together here. I think it's a good test to have, so I'll just leave this test as is and duplicate it to check if they work ok together

crates/swift-bridge-ir/src/bridged_type/shared_struct.rs
chinedufn: Thanks for explaining your thinking. Makes sense.

---

The issue is that we only want to support derives that we know are well behaved. 
Supporting arbitrary derives might be possible, but requires more research.

As it stands now, supporting arbitrary derives can lead to undefined behavior. 
It's incredibly unlikely, but theoretically possible. 

Before we support arbitrary derives we want to make UB *completely* impossible.

---

Essentially, `#[derive(SomeCustomDerive)]` could emit something like `type u32 = some_crate::NotAU32` and cause `swift-bridge` to think that it sees a `u32` not realizing that a type alias got generated to make `u32` a completely different type.

It should be possible to guard against this, but until we guard against it (and any other potential derive related UB scenarios) we can't allow arbitrary derives.

This discussion has more detail https://github.com/chinedufn/swift-bridge/issues/190#issuecomment-1462315921

rkreutz:
> See the following files in https://github.com/chinedufn/swift-bridge/pull/194/files for an example that you can copy/paste/modify

> crates/swift-bridge-ir/src/codegen/codegen_tests/derive_attribute_codegen_tests.rs

@chinedufn should I wait for that PR to be merged to add the tests on this file you mentioned, or should I create a separate file like `derive_struct_attribute_codegen_tests.rs`?

crates/swift-bridge-ir/src/bridged_type/shared_struct.rs
rkreutz: makes sense, will go with your suggestion then

crates/swift-bridge-ir/src/parse/parse_struct.rs
rkreutz: [ead0af9](https://github.com/chinedufn/swift-bridge/pull/198/commits/ead0af99013d5cafb5e2cb8d0da6ba7a191d8211)

crates/swift-bridge-ir/src/bridged_type/shared_struct.rs
rkreutz: [ead0af9](https://github.com/chinedufn/swift-bridge/pull/198/commits/ead0af99013d5cafb5e2cb8d0da6ba7a191d8211)

crates/swift-bridge-ir/src/codegen/generate_rust_tokens/shared_struct.rs
rkreutz: [ead0af9](https://github.com/chinedufn/swift-bridge/pull/198/commits/ead0af99013d5cafb5e2cb8d0da6ba7a191d8211)

crates/swift-bridge-ir/src/parse/parse_struct.rs
```diff
@@ -328,20 +343,48 @@ mod tests {
...
         let ty = module.types.types()[0].unwrap_shared_struct();
         
-        let derives: Vec<String> = ty.derives.clone().unwrap().iter().map(|derive| derive.to_string()).collect();
-        assert_eq!(derives, vec!["Copy", "Clone"]);
+        assert_eq!(ty.derives.copy, true);
+        assert_eq!(ty.derives.clone, true);
+
+        let ty2 = module.types.types()[1].unwrap_shared_struct();
+
+        assert_eq!(ty2.derives.copy, false);
```
chinedufn, line 361: nice

crates/swift-integration-tests/src/struct_attributes/derive.rs
```diff
@@ -0,0 +1,28 @@
+#[swift_bridge::bridge]
```
chinedufn, line 1: Awesome tests. Might've been nice to actually use the `Copy/Clone` impls outside of the module down below, but this is fine.

chinedufn (approved):
Looks great! Thanks!!

chinedufn:
Creating a file in this PR sounds good!

rkreutz:
> Creating a file in this PR sounds good!

just updated with the tests

chinedufn:
Looks good. I'll merge once tests are passing (looks like you need to run `cargo fmt`)

chinedufn:
Thank you!

extrawurst:
This would be great to have on `enum` too

PR #208, opened 2023-04-02
Generate include corresponding to each field in tuples

NiwakaDev:
This commit adds support for generating a `#include<~>` corresponding to each field in tuples. 

Here's an example.

```rust
#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        fn some_function(arg1: (i32, bool)) -> (i32, bool);
    }
}
```

Generates the following c header file:
```c
#include <stdint.h>
#include <stdbool.h>
//...
```

### DIFF
--- a/crates/swift-bridge-ir/src/bridged_type.rs
+++ b/crates/swift-bridge-ir/src/bridged_type.rs
@@ -118,7 +118,7 @@ pub(crate) trait BridgeableType: Debug {
 
     /// Generate a C include statement to put in the C header.
     /// For example, for a `u8` we would generate a `#include <stdint.h>` line.
-    fn to_c_include(&self) -> Option<&'static str>;
+    fn to_c_include(&self, types: &TypeDeclarations) -> Option<Vec<&'static str>>;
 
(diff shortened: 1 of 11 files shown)


NiwakaDev:
If this commit is merged, we can easily support generating `#include<~>` corresponding to each field in transparent structs or enums.

crates/swift-bridge-ir/src/bridged_type.rs
```diff
@@ -1645,9 +1645,9 @@ impl BridgedType {
         }
     }
 
-    pub fn to_c_include(&self) -> Option<&'static str> {
+    pub fn to_c_include(&self, types: &TypeDeclarations) -> Option<Vec<&'static str>> {
```
NiwakaDev, line 1648: Maybe, we could replace `Option<Vec<&'static str>>` with `Vec<&'static str>`.

chinedufn: I'm fine with whatever you decide.

chinedufn (approved):
Thanks!

PR #209, opened 2023-04-07
derive(Debug) enum fixes

naturecodevoid:
- fix: Add Debug to derived traits on Rust side even if one of the variants has data to hopefully avoid confusing the developer
- fix: don't generate Debug functions in C header and Swift if any variants have data to avoid confusing the developer with undefined symbol build errors

### DIFF
--- a/crates/swift-bridge-ir/src/codegen/generate_rust_tokens/shared_enum.rs
+++ b/crates/swift-bridge-ir/src/codegen/generate_rust_tokens/shared_enum.rs
@@ -143,19 +143,23 @@ impl SwiftBridgeModule {
         // We currently only allow derive(Debug) on non data carrying enums in order
         // to prevent a potential memory safety issue.
         // https://github.com/chinedufn/swift-bridge/pull/194#discussion_r1134386788
-        if shared_enum.derive.debug && !shared_enum.has_one_or_more_variants_with_data() {
+        if shared_enum.derive.debug {
+            // We don't want to confuse the developer if one of our variants has data and Debug isn't derived,
+            // so we still want to derive(Debug) on the Rust side.
+            // TODO: push a warning if one of our variants has data?
             derives.push(quote! {::std::fmt::Debug});
-
-            // __swift_bridge__$SomeEnum$Debug
-            let export_name = format!("{}$Debug", shared_enum.ffi_name_string());
-            // __swift_bridge__SomeEnum_Debug
-            let fn_name = format_ident!("{}_Debug", enum_ffi_name);
-            derive_impl_ffi_bridges.push(quote! {
-                #[export_name = #export_name]
-                pub extern "C" fn #fn_name(this: #enum_ffi_name) -> *mut swift_bridge::string::RustString {
-                    swift_bridge::string::RustString(format!("{:?}", this.into_rust_repr())).box_into_raw()
-                }
-            });
+            if !shared_enum.has_one_or_more_variants_with_data() {
+                // __swift_bridge__$SomeEnum$Debug
+                let export_name = format!("{}$Debug", shared_enum.ffi_name_string());
+                // __swift_bridge__SomeEnum_Debug
+                let fn_name = format_ident!("{}_Debug", enum_ffi_name);
+                derive_impl_ffi_bridges.push(quote! {
+                    #[export_name = #export_name]
+                    pub extern "C" fn #fn_name(this: #enum_ffi_name) -> *mut swift_bridge::string::RustString {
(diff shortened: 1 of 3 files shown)


chinedufn:
Hey, thanks for working on this.

Mind updating your PR to more clearly explain the problem and how this PR solves the problem, as well as to include an example bridge module that illustrates what this PR enables?

Thanks!

naturecodevoid:
Added!

naturecodevoid:
@chinedufn sorry to bother you, any chance of this being merged?

NiwakaDev:
Can we add some tests?

naturecodevoid:
What tests need to be added?

naturecodevoid:
@junepark678 You could add my branch as a dependency to use it early. I'll sync up the branch

crates/swift-bridge-ir/src/codegen/generate_rust_tokens/shared_enum.rs
```diff
@@ -140,22 +140,27 @@ impl SwiftBridgeModule {
         // User derives
         let mut derive_impl_ffi_bridges = vec![];
 
-        // We currently only allow derive(Debug) on non data carrying enums in order
-        // to prevent a potential memory safety issue.
-        // https://github.com/chinedufn/swift-bridge/pull/194#discussion_r1134386788
-        if shared_enum.derive.debug && !shared_enum.has_one_or_more_variants_with_data() {
+        if shared_enum.derive.debug {
+            // We don't want to confuse the developer if one of our variants has data and Debug isn't derived,
+            // so we still want to derive(Debug) on the Rust side.
```
chinedufn, line 145: You can remove this first part of the comment.

We don't need to explain why we always emit the `Debug` impl on the Rust side. This is desirable behavior.

Can keep the TODO. It's worth considering.

chinedufn (commented):
Here's how we can land this PR.

---

Update the PR body with an example of the "undefined symbol" error.

---

Add a `#[derive(Debug)]` section to the book that mentions that we currently only allow Swift to call the Rust Debug impl if the enum
is non-data-carrying, because we do not currently have a way to pass enums by reference (non-data-carrying enums are copied).

https://github.com/chinedufn/swift-bridge/pull/194#discussion_r1134386788

Can add the docs somewhere in here https://github.com/chinedufn/swift-bridge/blob/ecd3b974754e32fc951b8bc4098ecef84e974821/book/src/bridge-module/transparent-types/enums/README.md?plain=1#L43

---

Add a codegen test that covers the code that you've introduced.

Your codegen test can test that we do not emit C and Swift code for a data-carrying enum.

Can use this codegen test as inspiration:

https://github.com/chinedufn/swift-bridge/blob/53b93f47cc32d605377194a43f3164745cea6dd6/crates/swift-bridge-ir/src/codegen/codegen_tests/derive_attribute_codegen_tests.rs#L5-L68

Can use these to confirm that the Swift/C code does not contain any generated Debug code

https://github.com/chinedufn/swift-bridge/blob/636fa2774870c052a18f06d98ca5f6966c09bdd6/crates/swift-bridge-ir/src/codegen/codegen_tests.rs#L107-L108

---

Add an integration test.

Can add an data-carrying enum here: https://github.com/chinedufn/swift-bridge/blob/53b93f47cc32d605377194a43f3164745cea6dd6/crates/swift-integration-tests/src/enum_attributes/derive.rs#L1-L7

Something like:

```rust
#[swift_bridge::bridge]
mod ffi {
    #[derive(Debug)]
    enum DeriveDebugEnum {
        Variant,
    }

    #[derive(Debug)]
    enum DeriveDebugEnumDataCarrying {
        DataCarryingVariant(String),
    }
}

fn assert_implements_debug<T: Debug>() {}

/// Verify that the FFI enums implement `Debug`.
/// This confirms that `swift-bridge` is emitting the `#[derive(Debug)]` attribute on the final enum.
fn _test_implements_debug() {
    assert_implements_debug::<ffi::DeriveDebugEnum>();
    assert_implements_debug::<ffi::DeriveDebugEnumDataCarrying>();
}
```

amsam0:
Is there anyone interested in picking up this PR? I am not actively working on anything that depends on swift-bridge and finishing this up seems like a lot for something that was intended to be a quick fix for a bug.

As for the error message, it's just `ld` or the Swift compiler saying it can't find a symbol that some swift code needed.

NiwakaDev:
@chinedufn 

https://github.com/chinedufn/swift-bridge/pull/209#issuecomment-2290199332

May I take over this PR if no one addresses this by next week?

chinedufn:
Absolutely, feel free. Thanks.

PR #215, opened 2023-04-24
Update incorrect-return-type.stderr

NiwakaDev:
This commit fixes failing the UI test caused by updating rustc version 1.69.

If you want to know more detail, please see #213.


chinedufn:
Thanks!

PR #216, opened 2023-04-24
Add `RustVec<T>.as_ptr` method

aiongg:
This PR adds a method to the `Vectorizable` protocol:

```swift
static func vecOfSelfAsPtr(vecPtr: UnsafeMutableRawPointer) -> UnsafePointer<SelfRef>
```

My reason for adding this method was to be able to easily pass a serialized Protobuf object back and forth between Rust and Swift. Motivating example:

```rust
// rust

// Receives a serialized protobuf message and returns a serialized protobuf message
fn send_serialized_proto(bytes: &[u8]) -> Option<Vec<u8>> {
    // ...
}
```

Without `.as_ptr()`, it was necessary to return the raw pointer from Rust, which meant also returning the length separately, so using the Rust method was a bit less ergonomic on both sides. Here's how it looks now:

```swift
// swift

let message = My_Proto_Message()
// ...
guard let input = message.serializedData() else { ... }

guard let output: RustVec<UInt8>? = data.withUnsafeBytes {
    (ptr: UnsafeRawBufferPointer) -> RustVec<UInt8>? in
    let u8ptr = ptr.bindMemory(to: UInt8.self)
    return send_serialized_proto(u8ptr)
} else { ... }

do {
    let response = try My_Proto_Response.init(
        serializedData: Data(
            // The `.as_ptr()` method keeps this very straightforward
            bytes: output.as_ptr(),
            count: output.len()
        )
    )
} catch { ... }

// ...
```

### DIFF
--- a/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/VecTests.swift
+++ b/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/VecTests.swift
@@ -36,6 +36,12 @@ class VecTests: XCTestCase {
         vec.push(value: 222)
         XCTAssertEqual(vec.get(index: 1), 222)
     }
+    func testRustVecU8AsPtr() throws {
+        let vec = RustVec<UInt8>()
+        vec.push(value: 10)
+        let ptr = vec.as_ptr()
+        XCTAssertEqual(ptr.pointee, 10)
+    }
     func testRustVecU8Iterator() throws {
         let vec = RustVec<UInt8>()
         vec.push(value: 111)
--- a/crates/swift-bridge-build/src/generate_core.rs
+++ b/crates/swift-bridge-build/src/generate_core.rs
@@ -143,8 +143,6 @@ fn conform_to_vectorizable(swift_ty: &str, rust_ty: &str) -> String {
     format!(
         r#"
 extension {swift_ty}: Vectorizable {{
-    public typealias Elem = {swift_ty}
-
     public static func vecOfSelfNew() -> UnsafeMutableRawPointer {{
         __swift_bridge__$Vec_{rust_ty}$new()
     }}
@@ -184,8 +182,8 @@ extension {swift_ty}: Vectorizable {{
         }}
     }}
 
-    public static func vecOfSelfAsPtr(vecPtr: UnsafeMutableRawPointer) -> UnsafePointer<Elem> {{
-        UnsafePointer<Elem>(OpaquePointer(__swift_bridge__$Vec_{rust_ty}$as_ptr(vecPtr)))
+    public static func vecOfSelfAsPtr(vecPtr: UnsafeMutableRawPointer) -> UnsafePointer<Self> {{
+        UnsafePointer<Self>(OpaquePointer(__swift_bridge__$Vec_{rust_ty}$as_ptr(vecPtr)))
     }}
 
     public static func vecOfSelfLen(vecPtr: UnsafeMutableRawPointer) -> UInt {{
--- a/crates/swift-bridge-build/src/generate_core.rs
+++ b/crates/swift-bridge-build/src/generate_core.rs
@@ -143,6 +143,8 @@ fn conform_to_vectorizable(swift_ty: &str, rust_ty: &str) -> String {
     format!(
         r#"
 extension {swift_ty}: Vectorizable {{
+    public typealias Elem = {swift_ty}
+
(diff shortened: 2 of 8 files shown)


SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/VecTests.swift
```diff
@@ -36,6 +36,12 @@ class VecTests: XCTestCase {
         vec.push(value: 222)
         XCTAssertEqual(vec.get(index: 1), 222)
     }
+    func testRustVecU8AsPtr() throws {
+        let vec = RustVec<UInt8>()
+        vec.push(value: 10)
+        let ptr = vec.as_ptr()
+        XCTAssertEqual(ptr.pointee, 10)
+    }
```
aiongg, line 44: Basic test as described in the issue.

crates/swift-bridge-build/src/generate_core.rs
```diff
@@ -143,6 +143,8 @@ fn conform_to_vectorizable(swift_ty: &str, rust_ty: &str) -> String {
     format!(
         r#"
 extension {swift_ty}: Vectorizable {{
+    public typealias Elem = {swift_ty}
+
```
aiongg, line 147: Add the new `typealias` up top, and below the method implementation - it's the same procedure for all of these generated code segments.

crates/swift-bridge-build/src/generate_core/rust_string.swift
```diff
@@ -87,6 +87,10 @@ extension RustString: Vectorizable {
         }
     }
 
+    public static func vecOfSelfAsPtr(vecPtr: UnsafeMutableRawPointer) -> UnsafePointer<UInt8> {
+        UnsafePointer<UInt8>(OpaquePointer(__swift_bridge__$Vec_RustString$as_ptr(vecPtr)))
+    }
+
```
aiongg, line 93: Special handling for `RustString`, which takes a `UInt8` type parameter instead of the usual `Self`.

crates/swift-bridge-build/src/generate_core/rust_vec.swift
```diff
@@ -86,6 +91,7 @@ extension UnsafeBufferPointer {
 }
 
 public protocol Vectorizable {
+    associatedtype Elem
     associatedtype SelfRef
     associatedtype SelfRefMut
```
aiongg, line 96: The new `associatedtype Elem` for the `Vectorizable` protocol

chinedufn (changes requested):
`.as_ptr()` for a `RustVec<RustString>` should not return an `UnsafePointer<UInt8>`, but rather an `UnsafePointer<RustStringRef>`.

Consider the following Rust code:

```rust
let strings: Vec<String> = vec![];
let ptr: *const String = strings.as_ptr();
```

`Vec<String>.as_ptr()` returns a `*const String`.

Maybe you were accidentally looking at `String.as_ptr()`, which is unrelated to this PR.

---

So, I think we can remove the `Elem` associated type and instead just have the existing `SelfRef` associated type https://github.com/chinedufn/swift-bridge/blob/53ab502791dcc22f755524c91f01dd6c61d5dfb1/crates/swift-bridge-build/src/generate_core/rust_vec.swift#L95 , which from `RustVec<RustString>` is a `RustStringRef` https://github.com/chinedufn/swift-bridge/blob/53ab502791dcc22f755524c91f01dd6c61d5dfb1/crates/swift-bridge-build/src/generate_core/rust_string.swift#L72

aiongg:
Ok great I'll try it. I think I tried that the first time and encountered some issue, but if you think it should work then I must have done something wrong so I'll try again and update the PR if it works.

chinedufn:
Mmm I'm not certain but it seems directionally correct. Let me know if you run into an issue.

aiongg:
@chinedufn you were right, I was confusing `RustString` with `RustVec<RustString>`. Works fine with `SelfRef` as you described. Makes more sense now, too. I guess that's what I get for trying to do it at 2am last time 😅

chinedufn:
Thanks looks good.

All that's missing is to update your PR body with quick example code showing of what this PR now enables.

Here's an example from another PR that you can use as inspiration https://github.com/chinedufn/swift-bridge/pull/189

These examples get used in our release notes, and they help folks that are looking at old PRs understand exactly what the PR does.

Once you add an example (and perhaps clean up the info in your PR body that is no longer true) we can merge this!

aiongg:
@chinedufn done, please review.

PR #218, opened 2023-04-25
Fix returning Result<Vec<>, ...> from Rust

### LINKED ISSUE #217: Returning Result<Vec<OpaqueType>, OpaqueType> no longer compiles
Was trying to use master for some of the unreleased changes, and ran into this. `BridgedType::to_alpha_numeric_underscore_name()` does not support `Vec` (or `Option`).

jfaust:
Fixes #217

### DIFF
--- a/crates/swift-bridge-ir/src/bridged_type.rs
+++ b/crates/swift-bridge-ir/src/bridged_type.rs
@@ -481,6 +481,7 @@ impl BridgeableType for BridgedType {
 
     fn is_passed_via_pointer(&self) -> bool {
         match self {
+            BridgedType::StdLib(StdLibType::Vec(_)) => true,
             BridgedType::StdLib(_) => false,
             BridgedType::Foreign(_) => false,
             BridgedType::Bridgeable(ty) => ty.is_passed_via_pointer(),
--- a/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/ResultTests.swift
+++ b/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/ResultTests.swift
@@ -158,4 +158,21 @@ class ResultTests: XCTestCase {
             }
         }
     }
+
+    /// Verify that we can receive a Result<Vec<>, OpaqueRust> from Rust
+    func testSwiftCallRustResultVecUInt32Rust() throws {
+        let vec = try! rust_func_return_result_of_vec_u32()
+        XCTAssertEqual(vec.len(), 3)
+        for (i, value) in vec.enumerated() {
+            XCTAssertEqual(UInt32(i), value)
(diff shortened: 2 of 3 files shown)


NiwakaDev:
Thank you for your contribution!! I`m not the owner of this repository, but I left one minor feedback.

I think we need to add a codegen test. Please see an implementation guide of #217

crates/swift-bridge-ir/src/bridged_type.rs
```diff
@@ -481,6 +481,7 @@ impl BridgeableType for BridgedType {
 
     fn is_passed_via_pointer(&self) -> bool {
         match self {
+            BridgedType::StdLib(StdLibType::Vec(_)) => true,
```
NiwakaDev, line 484: If we use `Result<Vec<RustOpaqueType>, RustOpaqueType>`, we need to test that `__private__ResultPtrAndPtr` is generated, not a custom result type.

chinedufn (approved):
@NiwakaDev thanks a lot for reviewing! I'd love if you continued to help with reviewing PRs since at this point you have a LOT of experience with the internals.

---

For this particular PR, I don't think that we need a codegen test since we already have a bunch of codegen tests for the different flavors of `Result<T, E>`.

If we wanted to test whether or not the `PtrAndPtr` was used I think we'd just add a test where we parse different `Result<T, E>` signatures into `BuiltInResult` and assert whether or not a custom FFI representation is used.. Or something like that..
Don't think we need a codegen test for every possible Result signature, just the different "flavors" of Result.

So, I think that this PR is fine with just the integration test and can land.

Just lmk if you disagree.

---

@jfaust can you please update the PR body with an example bridge module illustrating what this PR enables (so, just a bridge module showing a function that returns a `Result<Vec<u32>, ()>` or something.

We like to have every PR show exactly what has been made possible.

Useful when we link new contributors to old PRs, as well as when we create release notes.

After that we can merge this.

Thanks!

crates/swift-bridge-ir/src/bridged_type.rs
chinedufn: As mentioned in my review I think this wouldn't need a codegen test but maybe instead we'd have some tests in the `BuiltInResult` file where we check which FFI representation is used by different `Result<T, E>` signatures.

So, we'd essentially be testing this function https://github.com/chinedufn/swift-bridge/blob/637c7b353539c1122dacc3a64e7d5dc052a4fec9/crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs#L395

Doesn't need to happen in this PR though.

Thanks a lot for reviewing this PR, very helpful!!! Thank you!

jfaust:
@chinedufn done!

chinedufn:
Thanks!

PR #221, opened 2023-04-25
async: Fix returning Result<(), OpaqueType>

### LINKED ISSUE #220: async: Error returning Result<(), OpaqueType>
I'm looking into this now, but figured I'd file this ticket while I do. This signature:
```
async fn rust_async_func_return_result_null() -> Result<(), AsyncResultOpaqueRustType2>;
```

Results in:
```
/Users/josh/work/easydrop/swift-bridge/SwiftRustIntegrationTestRunner/Generated/swift-integration-tests/swift-integration-tests.swift:3949:33: Cannot call value of non-function type '()'

/Users/josh/work/easydrop/swift-bridge/SwiftRustIntegrationTestRunner/Generated/swift-integration-tests/swift-integration-tests.swift:3963:73: Cannot convert value of type '(UnsafeMutableRawPointer?, __private__ResultPtrAndPtr) -> ()' to expected argument type '(@convention(c) (UnsafeMutableRawPointer?, UnsafeMutableRawPointer?) -> Void)?'
```
Screenshot 2023-04-25 at 9 58 15 AM

jfaust:
Fixes #220 

This supports bridges of the form:
```rust
#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        type ErrorType;
        async fn some_function() -> Result<(), ErrorType>;
    }
}

async fn some_function() -> Result<(), ErrorType> {
  Ok(())
}
```

### DIFF
--- a/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/AsyncTests.swift
+++ b/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/AsyncTests.swift
@@ -180,5 +180,9 @@ class AsyncTests: XCTestCase {
     func testSwiftCallsRustAsyncFnRetStruct() async throws {
         let _: AsyncRustFnReturnStruct = await rust_async_return_struct()
     }
+    
+    func testSwiftCallsRustAsyncFnReturnResultNullOpaqueRust() async throws {
+        try await rust_async_func_return_result_null_opaque_rust()
+    }
 }
 
--- a/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/AsyncTests.swift
+++ b/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/AsyncTests.swift
@@ -182,7 +182,14 @@ class AsyncTests: XCTestCase {
     }
     
     func testSwiftCallsRustAsyncFnReturnResultNullOpaqueRust() async throws {
-        try await rust_async_func_return_result_null_opaque_rust()
+        try await rust_async_func_return_result_null_opaque_rust(true)
+        
+        do {
+            try await rust_async_func_return_result_null_opaque_rust(false)
+            XCTFail()
+        } catch let error as AsyncResultOpaqueRustType2 {
+            XCTAssertEqual(error.val(), 111)
+        }
     }
 }
 
--- a/crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
+++ b/crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
@@ -190,6 +190,9 @@ impl BuiltInResult {
                 if self.is_custom_result_type() {
                     return format!("{}${}", SWIFT_BRIDGE_PREFIX, self.custom_c_struct_name());
                 }
+                if self.ok_ty.can_be_encoded_with_zero_bytes() {
+                    return "UnsafeMutableRawPointer?".to_string();
+                }
                 "__private__ResultPtrAndPtr".to_string()
             }
         }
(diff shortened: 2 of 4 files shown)


SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/AsyncTests.swift
```diff
@@ -180,5 +180,9 @@ class AsyncTests: XCTestCase {
     func testSwiftCallsRustAsyncFnRetStruct() async throws {
         let _: AsyncRustFnReturnStruct = await rust_async_return_struct()
     }
+    
+    func testSwiftCallsRustAsyncFnReturnResultNullOpaqueRust() async throws {
+        try await rust_async_func_return_result_null_opaque_rust()
+    }
```
chinedufn, line 186: Let's test both the ok and error case (I see that the other tests don't, but don't want to continue that habit).

One way would be to pass in a bool to `rust_async_func_return_result_null_opaque_rust` whether or not to return an error.

Then call it twice with `let _: () = try await  ...` and `let _: NameOfRustType = try await ...`

Can do this however you want though

crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
```diff
@@ -441,11 +444,25 @@ typedef struct {c_enum_name}{{{c_tag_name} tag; union {c_fields_name} payload;}}
...
+
         format!(
-            r#"if rustFnRetVal.is_ok {{
-        wrapper.cb(.success({ok}(ptr: rustFnRetVal.ok_or_err!)))
+            r#"if {condition} {{
+        wrapper.cb(.success({ok_val}))
     }} else {{
-        wrapper.cb(.failure({err}(ptr: rustFnRetVal.ok_or_err!)))
+        wrapper.cb(.failure({err_val}))
     }}"#
```
chinedufn, line 466: Let's add a codegen test since we're changing codegen.

Mainly because without a robust codegen test suite diving into future problems / implementing future optimizations would become a nightmare where a bunch of integration tests fail and it's hard to investigate why.

And especially since this is one of the most common Results that we'd expect users to use in the wild.

We can just add a `-> Result<(), OpaqueRustType>` codegen test to the Result codegen tests.

https://github.com/chinedufn/swift-bridge/blob/20baa731361c9cb86c2b4a9f5c7d0a5f65f6ff45/crates/swift-bridge-ir/src/codegen/codegen_tests/result_codegen_tests.rs#L6-L5

There are plenty of example async codegen tests here https://github.com/chinedufn/swift-bridge/blob/2d60d9ee83a153b07c5413af9f9cc3ad6075c9f4/crates/swift-bridge-ir/src/codegen/codegen_tests/async_function_codegen_tests.rs#L182-L268

chinedufn (commented):
Thanks a lot for whipping this up!!

Looks great. Left very minor feedback then this can land.

chinedufn:
Also, please update your PR body with an example bridge module illustrating what this PR has fixed.

jfaust:
@chinedufn both of those - the output from the test when it fails is stripped of whitespace, which makes it hard to parse what's different, even when doing a side by side diff.

But initially yes, knowing what to put for the Rust tokens. In my first PR a little while ago it was fairly easy, because all the changes to the output were ones I had explicitly made. With this PR, the actual change was tiny and caused large differences in output due to code paths I have no familiarity with. So I had to make the assumption that the code being generated was correct (it compiled and ran), and use the `rust-analyzer: Expand macro recursively` option in VSCode to find what was necessary to put into the test.

chinedufn:
Gotcha, thanks for explaining. Any suggestions for what could have made your contribution process easier?

chinedufn:
Hmm yeah we could probably use something like [`prettyplease`](https://github.com/dtolnay/prettyplease) for that, good idea.

That'd be great if you have the time.

Thanks for the ideas!

Issue #201, opened 2023-03-18
Various errors calling Swift from Rust


paholg:
Hello! I am trying to call Swift from Rust, but having a heck of a time getting anything but the most trivial example to work.

In all of the examples below, I have edited `src/main.rs` and `swift-library/Sources/swift-library/swift_library.swift` in the `rust-binary-calls-swift-package` example.

---

If I try to return an option, like so:

```rust
#[swift_bridge::bridge]
mod ffi {
    extern "Swift" {
        fn opt() -> Option<usize>;
    }
}
```
```swift
func opt() -> Optional<UInt> {
    Optional.none
}
```
I get the following error:

```
  /Users/paho/git/swift-bridge/examples/rust-binary-calls-swift-package/swift-library/Sources/swift-library/generated/rust-calls-swift/rust-calls-swift.swift:3:31: error: cannot convert return expression of type '__private__OptionUsize' to return type 'UInt'
```

as well as some warnings about dangling pointers. Looking at the generated code, the generated function is returning a plain `UInt`:

```
@_cdecl("__swift_bridge__$opt")
func __swift_bridge__opt () -> UInt {
    { let val = opt(); return __private__OptionUsize(val: val ?? 123, is_some: val != nil); }()
}
```

---

If I try to pass a `String` from Rust to Swift:

```rust
#[swift_bridge::bridge]
mod ffi {
    extern "Swift" {
        fn string(str: String);
    }
}
```
```swift
func string(str: RustString) {}
```

Then I get dangling pointer warnings still, and this error:

```
  /Users/paho/git/swift-bridge/examples/rust-binary-calls-swift-package/swift-library/Sources/swift-library/generated/rust-calls-swift/rust-calls-swift.swift:2:37: error: cannot find type 'GenericIntoRustString' in scope
  func __swift_bridge__string (_ str: GenericIntoRustString) {
                                      ^~~~~~~~~~~~~~~~~~~~~
```

---

Finally, if I try to pass in a `Vec` (replace `String` with `Vec<usize>` and `RustString` with `RustVec<UInt>` in the previous example), I get this error:

```
  /Users/paho/git/swift-bridge/examples/rust-binary-calls-swift-package/swift-library/Sources/swift-library/generated/rust-calls-swift/rust-calls-swift.swift:3:30: error: cannot convert value of type 'RustVec<UInt>' to expected argument type 'UnsafeMutableRawPointer'
      vector(vec: RustVec(ptr: vec))                          ^
```

---

[...]

Issue #170, opened 2023-02-15
Make nested struct declarations order independent


bes:
This code

```rust
#[swift_bridge::bridge]
mod ffi {
    #[swift_bridge(swift_repr = "struct")]
    pub struct TopLevel {
        pub nested: Option<Nested>,
    }

    #[swift_bridge(swift_repr = "struct")]
    pub struct Nested {
        x: f64,
    }
}
```

Won't compile in XCode since the generated C header files output the C structs in the wrong order.

`swift-bridge` outputs the structs in the order they are found inside a `#[swift_bridge::bridge]` block, but C requires `Nested` to always be defined before `TopLevel`.

Code output:

```c
typedef struct __swift_bridge__$TopLevel { struct __swift_bridge__$Option$Nested nested; } __swift_bridge__$TopLevel;
typedef struct __swift_bridge__$Nested { double x; } __swift_bridge__$Nested;
typedef struct __swift_bridge__$Option$Nested { bool x; __swift_bridge__$Nested val; } __swift_bridge__$Option$Nested;
```

Error message:

```
Field has incomplete type 'struct __swift_bridge__$Option$Nested'
```

I guess it either needs to be documented, or `swift-bridge` would need to sort the structs before generating C headers.

chinedufn:
Ah, thanks for opening this issue.

We want bridge modules to be type order independent.

Here's a step-by-step guide for anyone that might want to work on this issue:

### Integration test

We can add a `mod order_indepedent_declarations` that can be used to test scenarios like the one in the PR body.

https://github.com/chinedufn/swift-bridge/blob/3c0d00da33fdf19e201eb121c7e6e370897f8070/crates/swift-integration-tests/src/lib.rs#L4-L18

With contents along the lines of:

```rust
/// Verify that we can declare a struct that has a field that uses another struct
/// that gets declared later.
#[swift_bridge::bridge]
mod nested_struct {
    #[swift_bridge(swift_repr = "struct")]
	struct GrandparentStruct {
        inner: ParentStruct
	}

    #[swift_bridge(swift_repr = "struct")]
	struct ParentStruct {
        inner: ChildStruct
	}

    #[swift_bridge(swift_repr = "struct")]
	struct ChildStruct;
}

// ... In the future we can add more cases, such as for nested enums ...
```

### Codegen test

We can add a `mod order_indepedent_declarations_codegen_tests` to our codegen tests

[...]

PR #225, opened 2023-05-06
Support `RustString` in extern Swift functions

NiwakaDev:
Related to #201. This commit adds support `RustString`  in extern Swift functions.

```Swift
//Swift
func reflect_swift_string(arg: RustString) -> RustString {
    arg
}

```

```Rust
//Rust
#[swift_bridge::bridge]
mod ffi {
    extern "Swift" {
        fn reflect_swift_string(arg: String) -> String;
    }
}

let foo = "foo";
let string = ffi::reflect_swift_string(foo.to_string());
assert_eq!(string.len(), 3);
assert_eq!(&string, foo);
```

### DIFF
--- a/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunner/String.swift
+++ b/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunner/String.swift
@@ -10,3 +10,7 @@ import Foundation
 func create_swift_string() -> String {
     "hello"
 }
+
+func reflect_swift_string(arg: RustString) -> RustString {
+    arg
+}
--- a/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunner/String.swift
+++ b/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunner/String.swift
@@ -11,6 +11,6 @@ func create_swift_string() -> String {
     "hello"
 }
 
-func reflect_swift_string(arg: RustString) -> RustString {
+func reflect_rust_string(arg: RustString) -> RustString {
     arg
 }
--- a/crates/swift-bridge-ir/src/codegen/codegen_tests/string_codegen_tests.rs
+++ b/crates/swift-bridge-ir/src/codegen/codegen_tests/string_codegen_tests.rs
@@ -262,3 +262,50 @@ func __swift_bridge__some_function () -> UnsafeMutableRawPointer {
         .test();
     }
 }
+
+/// Test code generation for Swift function that takes and returns an owned String argument.
+mod extern_swift_func_takes_and_returns_string {
+    use super::*;
+
+    fn bridge_module_tokens() -> TokenStream {
+        quote! {
+            mod foo {
+                extern "Swift" {
+                    fn some_function (value: String) -> String;
+                }
+            }
+        }
+    }
+
+    fn expected_rust_tokens() -> ExpectedRustTokens {
+        ExpectedRustTokens::Contains(quote! {
+            pub fn some_function (value: String) -> String {
+                unsafe {
+                    Box :: from_raw (unsafe { __swift_bridge__some_function (swift_bridge :: string :: RustString (value) . box_into_raw ()) }) . 0
+                }
+            }
+        })
+    }
+
+    const EXPECTED_SWIFT_CODE: ExpectedSwiftCode = ExpectedSwiftCode::ContainsAfterTrim(
+        r#"
+@_cdecl("__swift_bridge__$some_function")
+func __swift_bridge__some_function (_ value: UnsafeMutableRawPointer) -> UnsafeMutableRawPointer {
+    { let rustString = some_function(value: RustString(ptr: value)).intoRustString(); rustString.isOwned = false; return rustString.ptr }()
+}
+"#,
+    );
+
+    const EXPECTED_C_HEADER: ExpectedCHeader = ExpectedCHeader::ExactAfterTrim(r#""#);
+
+    #[test]
+    fn extern_rust_fn_takes_and_returns_string() {
+        CodegenTest {
+            bridge_module: bridge_module_tokens().into(),
+            expected_rust_tokens: expected_rust_tokens(),
+            expected_swift_code: EXPECTED_SWIFT_CODE,
+            expected_c_header: EXPECTED_C_HEADER,
+        }
+        .test();
+    }
+}
--- a/crates/swift-integration-tests/src/string.rs
+++ b/crates/swift-integration-tests/src/string.rs
@@ -8,7 +8,7 @@ mod ffi {
 
     extern "Swift" {
         fn create_swift_string() -> String;
-        fn reflect_swift_string(arg: String) -> String;
+        fn reflect_rust_string(arg: String) -> String;
     }
 }
 
@@ -18,7 +18,7 @@ fn run_string_tests() {
     assert_eq!(&string, "hello");
 
     let foo = "foo";
-    let string = ffi::reflect_swift_string(foo.to_string());
+    let string = ffi::reflect_rust_string(foo.to_string());
     assert_eq!(string.len(), 3);
     assert_eq!(&string, foo);
 }
--- a/crates/swift-integration-tests/src/string.rs
+++ b/crates/swift-integration-tests/src/string.rs
@@ -8,6 +8,7 @@ mod ffi {
 
     extern "Swift" {
         fn create_swift_string() -> String;
+        fn reflect_swift_string(arg: String) -> String;
     }
 }
 
@@ -15,6 +16,11 @@ fn run_string_tests() {
     let string = ffi::create_swift_string();
     assert_eq!(string.len(), 5);
     assert_eq!(&string, "hello");
+
+    let foo = "foo";
+    let string = ffi::reflect_swift_string(foo.to_string());
+    assert_eq!(string.len(), 3);
+    assert_eq!(&string, foo);
 }
 
 fn create_string(str: &str) -> String {
--- a/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/StringTests.swift
+++ b/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/StringTests.swift
@@ -40,7 +40,7 @@ class StringTests: XCTestCase {
     
     func testRustStringToString() throws {
         let string = "hi"
-        
+
         XCTAssertEqual(
             create_string(string).toString(),
             "hi"
--- a/crates/swift-bridge-ir/src/bridged_type/bridgeable_string.rs
+++ b/crates/swift-bridge-ir/src/bridged_type/bridgeable_string.rs
@@ -49,7 +49,13 @@ impl BridgeableType for BridgedString {
 
     fn to_swift_type(&self, type_pos: TypePosition, _types: &TypeDeclarations) -> String {
         match type_pos {
-            TypePosition::FnArg(_func_host_lang, _) => "GenericIntoRustString".to_string(),
+            TypePosition::FnArg(func_host_lang, _) => {
+                if func_host_lang.is_rust() {
+                    "GenericIntoRustString".to_string()
+                } else {
+                    "UnsafeMutableRawPointer".to_string()
+                }
+            }
             TypePosition::FnReturn(func_host_lang) => {
                 if func_host_lang.is_rust() {
                     "RustString".to_string()


crates/swift-integration-tests/src/string.rs
```diff
@@ -8,13 +8,19 @@ mod ffi {
 
     extern "Swift" {
         fn create_swift_string() -> String;
+        fn take_and_return_swift_string(arg: String) -> String;
     }
 }
 
 fn run_string_tests() {
     let string = ffi::create_swift_string();
```
NiwakaDev, line 16: Maybe, if this PR get merged, we could remove this from `crates/swift-integration-tests/src/string.rs`.

crates/swift-bridge-ir/src/codegen/codegen_tests/string_codegen_tests.rs
```diff
@@ -262,3 +262,50 @@ func __swift_bridge__some_function () -> UnsafeMutableRawPointer {
...
+                }
+            }
+        })
+    }
+
+    const EXPECTED_SWIFT_CODE: ExpectedSwiftCode = ExpectedSwiftCode::ContainsAfterTrim(
+        r#"
+@_cdecl("__swift_bridge__$some_function")
+func __swift_bridge__some_function (_ value: UnsafeMutableRawPointer) -> UnsafeMutableRawPointer {
+    { let rustString = some_function(value: RustString(ptr: value)).intoRustString(); rustString.isOwned = false; return rustString.ptr }()
```
chinedufn, line 294: Why do we need to call `intoRustString` if, according to your PR body, `some_function` would return a `RustString` here?

crates/swift-integration-tests/src/string.rs
chinedufn: Makes sense

chinedufn (commented):
Left one question.

Tests are failing, but it seems like it started here https://github.com/chinedufn/swift-bridge/commit/7c14ea07ab71702019e652fcc5a5f7a111d5ea0e

The PR was passing when I merged it so maybe it conflicted with something in the master branch..

chinedufn:
Mind adding an example bridge module to your PR body so that it's more clear that the bridge module uses `String` and not `RustString`

NiwakaDev:
>Left one question.
>Tests are failing, but it seems like it started here https://github.com/chinedufn/swift-bridge/commit/7c14ea07ab71702019e652fcc5a5f7a111d5ea0e
>The PR was passing when I merged it so maybe it conflicted with something in the master branch..

By the way, Tests also failed before #218 got merged.

chinedufn:
Hmmm gotcha. I can't dive in for a couple days. Any idea where the issue might be stemming from?

NiwakaDev:
>Any idea where the issue might be stemming from?

I don't understand what causes the failure. I'll investigate it this week.

It might be a good idea to start by addressing the compiler warnings, like so:
```
/Users/runner/work/swift-bridge/swift-bridge/SwiftRustIntegrationTestRunner/Generated/swift-integration-tests/swift-integration-tests.h:741:232: warning: declaration of 'struct __swift_bridge__$ResultAsyncResultOkEnumAndAsyncResultOpaqueRustType1' will not be visible outside of this function
```

NiwakaDev:
I found that what cause the failure. I'll make a PR for fixing this problem tomorrow.

NiwakaDev:
I found that #213 causes the failure.

## Detail

Let's say that we have a tuple type and a result type, like so:

```rust
mod ffi {
    extern "Rust" {
        type ResultTestOpaqueRustType;
    }

    extern "Rust" {
        fn rust_func_return_result_tuple_transparent_enum(
            succeed: bool,
        ) -> Result<(i32, ResultTestOpaqueRustType, String), i32>;
    }
}
```

Will potentiality generate:

```C
struct __swift_bridge__$ResultTupleI32ResultTestOpaqueRustTypeStringAndI32 __swift_bridge__$rust_func_return_result_tuple_transparent_enum(bool succeed);
typedef enum __swift_bridge__$ResultTupleI32ResultTestOpaqueRustTypeStringAndI32$Tag {__swift_bridge__$ResultTupleI32ResultTestOpaqueRustTypeStringAndI32$ResultOk, __swift_bridge__$ResultTupleI32ResultTestOpaqueRustTypeStringAndI32$ResultErr} __swift_bridge__$ResultTupleI32ResultTestOpaqueRustTypeStringAndI32$Tag;
union __swift_bridge__$ResultTupleI32ResultTestOpaqueRustTypeStringAndI32$Fields {struct __swift_bridge__$tuple$I32ResultTestOpaqueRustTypeString ok; int32_t err;};
typedef struct __swift_bridge__$ResultTupleI32ResultTestOpaqueRustTypeStringAndI32{__swift_bridge__$ResultTupleI32ResultTestOpaqueRustTypeStringAndI32$Tag tag; union __swift_bridge__$ResultTupleI32ResultTestOpaqueRustTypeStringAndI32$Fields payload;} __swift_bridge__$ResultTupleI32ResultTestOpaqueRustTypeStringAndI32;
typedef struct __swift_bridge__$tuple$I32ResultTestOpaqueRustTypeString { int32_t _0; void* _1; void* _2; } __swift_bridge__$tuple$I32ResultTestOpaqueRustTypeString;           
```

Or 

```C
struct __swift_bridge__$ResultTupleI32ResultTestOpaqueRustTypeStringAndI32 __swift_bridge__$rust_func_return_result_tuple_transparent_enum(bool succeed);
typedef struct __swift_bridge__$tuple$I32ResultTestOpaqueRustTypeString { int32_t _0; void* _1; void* _2; } __swift_bridge__$tuple$I32ResultTestOpaqueRustTypeString;
typedef enum __swift_bridge__$ResultTupleI32ResultTestOpaqueRustTypeStringAndI32$Tag {__swift_bridge__$ResultTupleI32ResultTestOpaqueRustTypeStringAndI32$ResultOk, __swift_bridge__$ResultTupleI32ResultTestOpaqueRustTypeStringAndI32$ResultErr} __swift_bridge__$ResultTupleI32ResultTestOpaqueRustTypeStringAndI32$Tag;
union __swift_bridge__$ResultTupleI32ResultTestOpaqueRustTypeStringAndI32$Fields {struct __swift_bridge__$tuple$I32ResultTestOpaqueRustTypeString ok; int32_t err;};
typedef struct __swift_bridge__$ResultTupleI32ResultTestOpaqueRustTypeStringAndI32{__swift_bridge__$ResultTupleI32ResultTestOpaqueRustTypeStringAndI32$Tag tag; union __swift_bridge__$ResultTupleI32ResultTestOpaqueRustTypeStringAndI32$Fields payload;} __swift_bridge__$ResultTupleI32ResultTestOpaqueRustTypeStringAndI32;
```

The former causes the failure because `__swift_bridge__$ResultTupleI32ResultTestOpaqueRustTypeStringAndI32$Fields` couldn't know `__swift_bridge__$tuple$I32ResultTestOpaqueRustTypeString`, on the other hand the later doesn't cause the failure.

The reason for the non-deterministic ordering is due to `HashSet<String>`:

https://github.com/chinedufn/swift-bridge/blob/7c14ea07ab71702019e652fcc5a5f7a111d5ea0e/crates/swift-bridge-ir/src/codegen/generate_c_header.rs#L367-L370

And `Vec<TokenStream>` I introduced in #213:

https://github.com/chinedufn/swift-bridge/blob/7c14ea07ab71702019e652fcc5a5f7a111d5ea0e/crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs#L384

## Potential Solution

### 1
`__swift_bridge__$ResultTupleI32ResultTestOpaqueRustTypeStringAndI32$Fields` needs to know `__swift_bridge__$tuple$I32ResultTestOpaqueRustTypeString`, but `__swift_bridge__$tuple$I32ResultTestOpaqueRustTypeString` could be generated without any dependencies. This means that we need to think the order of generating the two declarations. Maybe by introducing  something like a directed graph or that to represent the type dependencies, we could fix this problem.

### 2

Since Solution 1 might be complicated, so maybe we could stop support #213 to fix the failure.

What do you think about this?

chinedufn:
Thanks for the detailed breakdown and links. This made it very easy for me to understand this complex issue.

I don't think we need a complex solution.

I think we can solve this with the approach described here https://github.com/chinedufn/swift-bridge/issues/170#issuecomment-1431588845

---

In essence, we use a `parse_queue: Vec<&TypeDeclaration>` and a `encountered: HashSet<String>`

We iterate over the types and push them to the parse queue. If the type contains other types we push those other types to the parse queue first, recursively.

Each time we push a type we insert its C header code into the `encountered` `HashSet`, this way we don't push the same type twice.

So, this guarantees that a type always appears after all of the types that we depend on.

---

So, because the solution should be fairly self-contained and not require much code, I like potential solution 1 the best where we just fix the problem.

---

Does that make sense?

crates/swift-bridge-ir/src/codegen/codegen_tests/string_codegen_tests.rs
chinedufn: @NiwakaDev waiting on this question then after that we can merge this and solve the test stuff in a separate PR

chinedufn: Oh and also please add an example bridge module to your PR body so that it's more clear that the bridge module uses String and not RustString

NiwakaDev:
Thank you for your quick reply.

>Each time we push a type we insert its C header code into the encountered HashSet, this way we don't push the same type twice.
So, this guarantees that a type always appears after all of the types that we depend on.

I guess that the contents order of `HashSet` isn't deterministic. Why does this guarantees it?.

NiwakaDev:
>Does that make sense?

Yes. Thanks!!

crates/swift-bridge-ir/src/codegen/codegen_tests/string_codegen_tests.rs
chinedufn: Any availability to look at this? One this is addressed I can merge

NiwakaDev: I am working on it, but I'm slightly busy this month. Sorry...

SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunner/String.swift
```diff
@@ -10,3 +10,7 @@ import Foundation
 func create_swift_string() -> String {
     "hello"
 }
+
+func take_and_return_swift_string(arg: RustString) -> String {
+    arg.toString()
+}
```
NiwakaDev, line 16: @chinedufn 

`create_swift_string` returns `String`. To maintain consistency and remove `toRustString`(`extern_swift_func_takes_and_returns_string`), I changed `take_and_return_swift_string` return value from `RustString` to `String`. But it is inefficient. what do you think about this?

NiwakaDev: Here're `create_swift_string` and `take_and_return_swift_string` implementation.
```String
//
//  String.swift
//  SwiftRustIntegrationTestRunner
//
//  Created by Frankie Nwafili on 2/18/22.
//

import Foundation

func create_swift_string() -> String {
    "hello"
}

func take_and_return_swift_string(arg: RustString) -> String {
    arg.toString()
}

```

chinedufn: I haven't reviewed the code yet but this sounds fine.

Let's just document this behavior in the `String` chapter in the book so that users are aware of this implicit clone

crates/swift-bridge-ir/src/codegen/codegen_tests/string_codegen_tests.rs
chinedufn: > Why do we need to call intoRustString if, according to your PR body, some_function would return a RustString here?

Nevermind, the reason is that we support returning anything that can be converted into a RustString.

crates/swift-integration-tests/src/string.rs
```diff
@@ -8,13 +8,19 @@ mod ffi {
 
     extern "Swift" {
         fn create_swift_string() -> String;
+        fn take_and_return_swift_string(arg: String) -> String;
```
chinedufn, line 11: Let's call this `take_and_return_rust_string` and make the Swift function `(arg: RustString) -> RustString`

This way we have a test for both `-> String` and `-> RustString`

chinedufn (approved):
Left minor feedback then looks good to me.

Can you

1. Address feedback
2. Update your PR body to show an example bridge module

After that you can merge this PR.

Thanks!

NiwakaDev:
@chinedufn 

Addressed your review. Please let me know if there are any errors or omissions.

chinedufn:
> Please let me know if there are any errors or omissions.

Minor, but the function should be called `reflect_rust_string` (not `reflect_swift_string`).

After that LGTM

NiwakaDev:
@chinedufn 

Done.

chinedufn:
Thanks!

PR #226, opened 2023-05-11
Fix `already_declared` attribute for enums and structs

conectado:
Hi! I encounted an issue while working with the crate while using `already_declared`:

When generating the rust function wrappers for an extern swift function with an already_declared struct/enum the generated code used a locally defined struct/enum with this patch we add `super` to that function signature.

An example of code that was previously not working:

```rs
#[swift_bridge::bridge]
mod ffi_i {
  enum Foo {
    Bar.
  }
}

use ffi_i::Foo;

#[swift_bridge::bridge]
mod ffi {
  #[swift_bridge(already_declared)]
  enum Foo {}
  
  extern "Rust" {}
  extern "Swift" {
    fn baz(a: Foo);
  }
}
```

Previously the genrated function would be something like:
```rs
mod ffi {
  fn baz(a: Foo) {
    // ...
  }
}
```

now it's

```rs
mod ffi {
  fn baz(a: super::Foo) {
    // ...
  }
}
```

I'm not sure if it fixes all edge cases but it's working for me locally.

Let me know if I there's something to change to get this merged or if I was just not using `already_declared` correctly 😅

### DIFF
--- a/crates/swift-integration-tests/src/enum_attributes/already_declared.rs
+++ b/crates/swift-integration-tests/src/enum_attributes/already_declared.rs
@@ -16,10 +16,18 @@ mod ffi2 {
     enum AlreadyDeclaredEnumTest {}
 
     extern "Rust" {
-        fn reflect_already_declared_enum(arg: AlreadyDeclaredEnumTest) -> AlreadyDeclaredEnumTest;
+        fn rust_reflect_already_declared_enum(
+            arg: AlreadyDeclaredEnumTest,
+        ) -> AlreadyDeclaredEnumTest;
+    }
+
+    extern "Swift" {
+        fn swift_reflect_already_declared_enum(
+            arg: AlreadyDeclaredEnumTest,
+        ) -> AlreadyDeclaredEnumTest;
     }
 }
 
-fn reflect_already_declared_enum(arg: AlreadyDeclaredEnumTest) -> AlreadyDeclaredEnumTest {
+fn rust_reflect_already_declared_enum(arg: AlreadyDeclaredEnumTest) -> AlreadyDeclaredEnumTest {
     arg
 }
--- a/crates/swift-integration-tests/src/enum_attributes/already_declared.rs
+++ b/crates/swift-integration-tests/src/enum_attributes/already_declared.rs
@@ -21,6 +21,10 @@ mod ffi2 {
         ) -> AlreadyDeclaredEnumTest;
     }
 
+    extern "Rust" {
+        fn test_rust_calls_swift_already_declared();
+    }
+
     extern "Swift" {
         fn swift_reflect_already_declared_enum(
             arg: AlreadyDeclaredEnumTest,
@@ -30,4 +34,8 @@ mod ffi2 {
 
 fn rust_reflect_already_declared_enum(arg: AlreadyDeclaredEnumTest) -> AlreadyDeclaredEnumTest {
     arg
+}
+fn test_rust_calls_swift_already_declared() {
+    let val = ffi2::swift_reflect_already_declared_enum(AlreadyDeclaredEnumTest::Variant);
+    assert!(matches!(val, AlreadyDeclaredEnumTest::Variant));
 }
--- a/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunner/SharedEnumAttributes.swift
+++ b/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunner/SharedEnumAttributes.swift
@@ -10,3 +10,7 @@ import Foundation
 func extern_swift_enum_rename(arg: EnumRename) -> EnumRename {
     arg
 }
+
+func swift_reflect_already_declared_enum(arg: AlreadyDeclaredEnumTest) -> AlreadyDeclaredEnumTest {
+    arg
+}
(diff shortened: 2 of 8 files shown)


chinedufn (commented):
Thanks for your first contribution!

Just need to add tests and then we can land this.

---

Can add an `extern "Swift"` function in here https://github.com/chinedufn/swift-bridge/blob/72e1759e0504e3a5a32587605c98e410f12ea9e7/crates/swift-integration-tests/src/enum_attributes/already_declared.rs#L13-L21

And in here https://github.com/chinedufn/swift-bridge/blob/72e1759e0504e3a5a32587605c98e410f12ea9e7/crates/swift-integration-tests/src/struct_attributes/already_declared.rs#L14-L24

---

And then you can add an extern "Swift" function to this codegen test https://github.com/chinedufn/swift-bridge/blob/2bffcb08c1f0ca8bbcba9c78c5ae28a9cddfb632/crates/swift-bridge-ir/src/codegen/codegen_tests/already_declared_attribute_codegen_tests.rs#L149-L177

and this one https://github.com/chinedufn/swift-bridge/blob/2bffcb08c1f0ca8bbcba9c78c5ae28a9cddfb632/crates/swift-bridge-ir/src/codegen/codegen_tests/already_declared_attribute_codegen_tests.rs#L102-L123

In both of the tests you can use the `ExpectedRustTokens::ContainsManyAndDoesNotContainMany` to confirm that we generate  code for the `extern Swift` function that uses `super::TypeName`

conectado:
> Thanks for your first contribution!
> 
> Just need to add tests and then we can land this.
> 
> Can add an `extern "Swift"` function in here
> 
> https://github.com/chinedufn/swift-bridge/blob/72e1759e0504e3a5a32587605c98e410f12ea9e7/crates/swift-integration-tests/src/enum_attributes/already_declared.rs#L13-L21
> 
> And in here
> 
> https://github.com/chinedufn/swift-bridge/blob/72e1759e0504e3a5a32587605c98e410f12ea9e7/crates/swift-integration-tests/src/struct_attributes/already_declared.rs#L14-L24
> 
> And then you can add an extern "Swift" function to this codegen test
> 
> https://github.com/chinedufn/swift-bridge/blob/2bffcb08c1f0ca8bbcba9c78c5ae28a9cddfb632/crates/swift-bridge-ir/src/codegen/codegen_tests/already_declared_attribute_codegen_tests.rs#L149-L177
> 
> and this one
> 
> https://github.com/chinedufn/swift-bridge/blob/2bffcb08c1f0ca8bbcba9c78c5ae28a9cddfb632/crates/swift-bridge-ir/src/codegen/codegen_tests/already_declared_attribute_codegen_tests.rs#L102-L123
> 
> In both of the tests you can use the `ExpectedRustTokens::ContainsManyAndDoesNotContainMany` to confirm that we generate code for the `extern Swift` function that uses `super::TypeName`

Thanks for the review and detailed guidance!

Added the tests :)

chinedufn (approved):
Nice!

crates/swift-integration-tests/src/enum_attributes/already_declared.rs
```diff
@@ -16,10 +16,18 @@ mod ffi2 {
...
-        fn reflect_already_declared_enum(arg: AlreadyDeclaredEnumTest) -> AlreadyDeclaredEnumTest;
+        fn rust_reflect_already_declared_enum(
+            arg: AlreadyDeclaredEnumTest,
+        ) -> AlreadyDeclaredEnumTest;
+    }
+
+    extern "Swift" {
+        fn swift_reflect_already_declared_enum(
+            arg: AlreadyDeclaredEnumTest,
+        ) -> AlreadyDeclaredEnumTest;
```
NiwakaDev, line 27: @chinedufn 

I guess that this function isn't called from anywhere.

chinedufn: Ah yeah we should call it. Great point. Want to tell @conectado how to best call it?

conectado: I added a test [here](https://github.com/chinedufn/swift-bridge/pull/226/commits/7415b9108beb87de95cd4687c1f841ae7b37ea6f) that I think does this.

I couldn't run the tests because I think the CI is failing so lmk if this work or I should do something else?

NiwakaDev: Hi, @conectado 
The CI failure is due to #213. If you want to know more detail, please see #225.

I guess it should be fine as long as you run the test multiple times and then the test pass at least once.

conectado: Locally it's consistently failing with:

> Testing failed:
>         field has incomplete type 'struct __swift_bridge__$tuple$I32ResultTestOpaqueRustTypeString'
>         failed to emit precompiled header '/Users/conectado/Library/Developer/Xcode/DerivedData/SwiftRustIntegrationTestRunner-cocipvlohdpefifldegpkicaiuld/Build/Intermediates.noindex/PrecompiledHeaders/BridgingHeader-swift_4X3E80DLUM90-clang_KV4PBZK8M8GA.pch' for bridging header '/Users/conectado/repos/firezone/swift-bridge/SwiftRustIntegrationTestRunner/Headers/BridgingHeader.h'
>         Testing cancelled because the build failed

NiwakaDev: This error is due to #213.

>Locally it's consistently failing with:

It is not consistently. Right now, when we run `./test-swift-rust-integration.sh`, The tests will sometimes fails. If you check #225, You could understand why the tests sometimes fails.

On the other hand, when we run `cargo test --all`, the tests will consistently pass.

chinedufn:
Yeah CI is currently failing due to https://github.com/chinedufn/swift-bridge/pull/225#issuecomment-1552361044

Test looks good, I'll land this. Thanks!

PR #229, opened 2023-06-14
Swift `Vec<primitive>` Return Type and Function Argument

### LINKED ISSUE #228: How to return `Vec<u8>` from Swift to Rust
I've been struggling with this for a while now, so now I'm reaching out for some help as I'm seemingly unable to figure it out by myself.

What I want to do is effectively the inverse of https://github.com/chinedufn/swift-bridge/issues/214

My Rust code will call into Swift with a couple parameters and once it completes, the Swift code should return a `Vec<u8>` of bytes back to my Rust code.

MWE is reasonably simple:
```rust
// Rust
fn main() {
    let vec = ffi:get_bytes();
    println!("vec: {:?}", vec);
}

#[swift_bridge::bridge]
mod ffi {
    extern "Swift" {
        fn get_bytes() -> Vec<u8>;
    }
}
```

```swift
// Swift
func get_bytes() -> RustVec<UInt8> {
    let vec = RustVec<UInt8>()
    // do some work to fill the Vec here
    return vec
}
```

This is what the codegen outputs:
```swift
// Swift
@_cdecl("__swift_bridge__$get_bytes")
func __swift_bridge__get_bytes () -> RustVec<UInt8> {
    { let val = get_bytes(); val.isOwned = false; return val.ptr }()
}
```

During compilation, this then throws an error at me:
`error: cannot convert value of type 'UnsafeMutableRawPointer' to closure result type 'RustVec<UInt8>`

Which makes sense, given that it returns `val.ptr`, which is not of type `RustVec<UInt8>`.

I'm very new to pointer wrangling and working with lower level stuff in between Rust/Swift, so there may already be a solution out there, however I was not able to find a working way forward.
I already tried some pointer shenanigans, but the codegen always seems to output return types that don't fit whatever I can give it, so I'm a little stuck.

TL;DR: How do I return a bucket of bytes from Swift to Rust?

timwedde:
Fixes #228.

This enables the following use case:
```rust
// Rust
fn main() {
    let bytes = ffi::receive_bytes();
    println!("bytes: {:?}", bytes);

    let vec: Vec<u8> = vec![6, 1, 2, 3, 4, 5];
    ffi::send_bytes(vec);
}

#[swift_bridge::bridge]
mod ffi {
    extern "Swift" {
        fn receive_bytes() -> Vec<u8>;
        fn send_bytes(vec: Vec<u8>);
    }
}
```

```swift
// Swift
func send_bytes(vec: RustVec<UInt8>) {
    print("Received \(vec.len()) bytes from Rust")
    for val in vec {
        print(val)
    }
}

func receive_bytes() -> RustVec<UInt8> {
    let vec = RustVec<UInt8>()
    for i in 0 ... 4 {
        vec.push(value: UInt8(i))
    }
    return vec
}
```

This is chiefly intended for easy bidirectional byte exchanges but should work for all primitive types, at least the ones I tested with.

The codegen tests pass but I was unable to get the integration tests to run: They also fail for me on `master` and I don't understand enough of how they work to properly debug them. The (_very_ long) error messages ends with this, which seems to be the culprit:
> field has incomplete type 'struct __swift_bridge__$tuple$I32ResultTestOpaqueRustTypeString'

I attempted to add some integration tests for both directions anyway, but given that I'm operating 'blind' on that part I can't assure you whether these work or not.

I'm rather certain that there's still issues with my PR given that this is my first time dealing with this kind of interop work, so by all means do let me know how to address the current weaknesses/problems and I'll do my best to resolve them.

### DIFF
--- a/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunner/Vec.swift
+++ b/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunner/Vec.swift
@@ -1,6 +1,12 @@
 import Foundation
 
-func swift_arg_vec_u8(vec _: RustVec<UInt8>) {}
+func swift_arg_vec_u8(vec _: RustVec<UInt8>) {
+    assert(vec[0] == 1)
+    assert(vec[1] == 2)
+    assert(vec[2] == 3)
+    assert(vec[3] == 4)
+    assert(vec[4] == 5)
+}
 
 func swift_return_vec_u8() -> RustVec<UInt8> {
     let vec = RustVec<UInt8>()
--- a/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunner/Vec.swift
+++ b/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunner/Vec.swift
@@ -1,10 +1,3 @@
-//
-//  String.swift
-//  SwiftRustIntegrationTestRunner
-//
-//  Created by Frankie Nwafili on 2/18/22.
-//
-
 import Foundation
 
 func send_bytes(vec _: RustVec<UInt8>) {}
--- a/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunner/Vec.swift
+++ b/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunner/Vec.swift
@@ -1,0 +1,18 @@
+//
+//  String.swift
+//  SwiftRustIntegrationTestRunner
+//
+//  Created by Frankie Nwafili on 2/18/22.
+//
+
+import Foundation
+
+func send_bytes(vec _: RustVec<UInt8>) {}
+
+func receive_bytes() -> RustVec<UInt8> {
+    let vec = RustVec<UInt8>()
+    for i in 0 ... 4 {
+        vec.push(value: UInt8(i))
+    }
+    return vec
+}
--- a/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunner/Vec.swift
+++ b/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunner/Vec.swift
@@ -1,8 +1,8 @@
 import Foundation
 
-func send_bytes(vec _: RustVec<UInt8>) {}
+func swift_arg_vec_u8(vec _: RustVec<UInt8>) {}
 
-func receive_bytes() -> RustVec<UInt8> {
+func swift_return_vec_u8() -> RustVec<UInt8> {
     let vec = RustVec<UInt8>()
     for i in 0 ... 4 {
         vec.push(value: UInt8(i))
--- a/crates/swift-bridge-ir/src/bridged_type.rs
+++ b/crates/swift-bridge-ir/src/bridged_type.rs
@@ -1158,9 +1158,25 @@ impl BridgedType {
                         unimplemented!()
                     }
                 },
-                StdLibType::Vec(ty) => {
-                    format!("RustVec<{}>", ty.ty.to_swift_type(type_pos, types))
-                }
+                StdLibType::Vec(ty) => match type_pos {
+                    TypePosition::FnArg(func_host_lang, _) => {
+                        if func_host_lang.is_rust() {
+                            format!("RustVec<{}>", ty.ty.to_swift_type(type_pos, types))
+                        } else {
+                            "UnsafeMutableRawPointer".to_string()
+                        }
+                    },
+                    TypePosition::FnReturn(func_host_lang) => {
+                        if func_host_lang.is_rust() {
+                            format!("RustVec<{}>", ty.ty.to_swift_type(type_pos, types))
+                        } else {
+                            "UnsafeMutableRawPointer".to_string()
+                        }
+                    }
+                    _ => {
+                        format!("RustVec<{}>", ty.ty.to_swift_type(type_pos, types))
+                    }
+                },
                 StdLibType::Option(opt) => opt.to_swift_type(type_pos, types),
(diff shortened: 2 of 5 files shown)


NiwakaDev:
@timwedde 
>field has incomplete type 'struct swift_bridge$tuple$I32ResultTestOpaqueRustTypeString'

This error is not due to this PR. If you want know more detail, please see ##225.

It seems like this PR is fine. Thanks!!

SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunner/Vec.swift
```diff
@@ -0,0 +1,18 @@
+//
+//  String.swift
+//  SwiftRustIntegrationTestRunner
+//
+//  Created by Frankie Nwafili on 2/18/22.
+//
```
chinedufn, line 6: You can delete this comment. Xcode typically auto generates them but we don't really need them

crates/swift-bridge-ir/src/codegen/codegen_tests/vec_codegen_tests.rs
```diff
@@ -544,3 +544,116 @@ void __swift_bridge__$some_function(void* arg);
...
+@_cdecl("__swift_bridge__$some_function")
+func __swift_bridge__some_function () -> UnsafeMutableRawPointer {
+    { let val = some_function(); val.isOwned = false; return val.ptr }()
+}
+"#,
+        )
+    }
+
+    fn expected_c_header() -> ExpectedCHeader {
+        ExpectedCHeader::ContainsAfterTrim("")
```
chinedufn, line 589: This will pass for all strings string all strings contain `""`. 

You can use `ExpectedCHeader::ExactAfterTrim("")` here instead.

SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunner/Vec.swift
```diff
@@ -0,0 +1,18 @@
...
+
+import Foundation
+
+func send_bytes(vec _: RustVec<UInt8>) {}
```
chinedufn, line 10: Lets call these `swift_arg_vec_u8` and `swift_return_vec_u8`.

This way we can more easily distinguish these from other FFI functions.

chinedufn, line 10: Let's make this function assert that the bytes are `1, 2, 3, 4, 5`

crates/swift-integration-tests/src/vec.rs
```diff
@@ -25,6 +25,28 @@ mod ffi {
...
+        fn run_vec_tests();
+    }
+
+    extern "Swift" {
+        fn receive_bytes() -> Vec<u8>;
+        fn send_bytes(vec: Vec<u8>);
+    }
+}
+
+fn run_vec_tests() {
```
chinedufn, line 39: Nice

crates/swift-bridge-ir/src/codegen/codegen_tests/vec_codegen_tests.rs
```diff
@@ -544,3 +544,116 @@ void __swift_bridge__$some_function(void* arg);
...
+@_cdecl("__swift_bridge__$some_function")
+func __swift_bridge__some_function (_ arg: UnsafeMutableRawPointer) {
+    some_function(arg: RustVec(ptr: arg))
+}
+"#,
+        )
+    }
+
+    fn expected_c_header() -> ExpectedCHeader {
+        ExpectedCHeader::ContainsAfterTrim("")
```
chinedufn, line 646: `::ExactAfterTrim`

crates/swift-integration-tests/src/vec.rs
```diff
@@ -25,6 +25,28 @@ mod ffi {
             arg: Vec<TransparentEnumInsideVecT>,
         ) -> Vec<TransparentEnumInsideVecT>;
     }
+
+    extern "Rust" {
+        fn run_vec_tests();
```
chinedufn, line 30: We need to call this on the Swift side in `VecTests.rs`

chinedufn: https://github.com/chinedufn/swift-bridge/blob/master/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/VecTests.swift#L11

something like:

```swift
/// Verify that Rust can pass `RustVec`s to and receive `RustVec`s from Swift.
func testRustCallsSwiftRustVecFunctions() {
    run_vec_tests()
}
```

chinedufn (commented):
Thanks for the very detailed PR body.

This looks great to me. Left some very minor feedback then after that this is good to go.

crates/swift-bridge-ir/src/bridged_type.rs
```diff
@@ -1158,9 +1158,25 @@ impl BridgedType {
...
-                    format!("RustVec<{}>", ty.ty.to_swift_type(type_pos, types))
-                }
+                StdLibType::Vec(ty) => match type_pos {
+                    TypePosition::FnArg(func_host_lang, _) => {
+                        if func_host_lang.is_rust() {
+                            format!("RustVec<{}>", ty.ty.to_swift_type(type_pos, types))
+                        } else {
+                            "UnsafeMutableRawPointer".to_string()
+                        }
+                    },
```
chinedufn, line 1168: Need to run `cargo fmt` https://github.com/chinedufn/swift-bridge/actions/runs/5272573390/jobs/9561689835

timwedde:
Alright, I think I addressed all of the comments, let me know how I did 👍

chinedufn:
Thanks!

chinedufn (approved):

PR #232, opened 2023-06-21
Recreate Vec.swift

NiwakaDev:
Before this commit, the integration tests consistently fail becasuse Xcode couldn't recognize `Vec.swift`, like so:

```shell
Testing failed:
        Cannot find 'swift_return_vec_u8' in scope
        Cannot find 'swift_arg_vec_u8' in scope
        Generic parameter 'T' could not be inferred
```

Vec

As you can see the above image, there isn't `Vec.swift` in the Xcode.

So, this commit recreates `Vec.swift` in the Xcode.


chinedufn:
Ah I just fixed it before seeing this PR so I'll close this. Thanks! https://github.com/chinedufn/swift-bridge/commit/df91f282e62b32755f1410c0e6f22b03f836f018

PR #244, opened 2023-09-10
Improve the error message in SharedStructDeclarationParser when an un…

bes:
…supported attribute is found

The panic message now contains the name of the unsupported attribute as well as the name of the struct that the attribute was found on.

This commit also updates the documentation to fix faulty struct attributes.


bes:
I just got bit by this yesterday, when upgrading swift-bridge from 0.1.49 to latest.

I had used `#[swift_bridge::bridge(swift_repr = "struct")]` on a struct, but that syntax is unsupported.

The correct syntax is `#[swift_bridge(swift_repr = "struct")]`.

chinedufn:
Thanks!

PR #250, opened 2024-01-28
added missing source

ItamarMu:
in build-swiftc-links-rust.sh, the command
swiftc -L target/x86_64-apple-darwin/debug/ -lswift_and_rust -import-objc-header bridging-header.h \
  main.swift lib.swift ./generated/swift-and-rust/swift-and-rust.swift 
is missing the source ./generated/SwiftBridgeCore.swift


chinedufn:
Thanks!

PR #203, opened 2023-03-18
Support Option<Primitive> in Swift fns

chinedufn:
This commit adds support for Option<PrimitiveType> in `extern "Swift"`
functions.

For example, the following is now possible:

```rust
mod ffi {
    extern "Swift" {
        fn my_func(arg: Option<u8>) -> Option<bool>;
    }
}
```

PR #257, opened 2024-03-14
Add test cases for Option<&T> and fix rust codegen

PrismaPhonic:
Add test cases for Option<&T> and fix rust codegen

Currently swift-bridge only has tests for Option<T> - this adds test
cases for Option<&T> and fixes a bug in rust codegen that does not
correctly translate Option<&T>.

This is now possible:

```rust
mod ffi {
  extern "Rust" {
    type MyStruct;

    fn my_func(arg: Option<&MyStruct>) -> Option<&MyStruct>;
  }
}
```

### DIFF
--- a/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/OptionTests.swift
+++ b/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/OptionTests.swift
@@ -106,13 +106,13 @@ class OptionTests: XCTestCase {
         let val = OptTestOpaqueRefRustType.new(123)
         let opt_ref = val.field_ref()
 
-        let reflect = rust_reflect_option_ref_opaque_rust_type(opt_ref)
+        var reflect = rust_reflect_option_ref_opaque_rust_type(opt_ref)
         XCTAssertEqual(reflect!.field(), 123)
         XCTAssertNil(rust_reflect_option_ref_opaque_rust_type(nil))
-        let reflect = nil
+        reflect = nil
         
-        let second_reference = rust_reflect_option_ref_opaque_rust_type(opt_ref)
-        XCTAssertEqual(second_reference!.field(), 123)
+        reflect = rust_reflect_option_ref_opaque_rust_type(opt_ref)
+        XCTAssertEqual(reflect!.field(), 123)
     }
     
     func testSwiftCallRustWithOptionOpaqueRustCopyType() throws {
--- a/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/OptionTests.swift
+++ b/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/OptionTests.swift
@@ -97,6 +97,23 @@ class OptionTests: XCTestCase {
         
         XCTAssertNil(rust_reflect_option_opaque_rust_type(nil))
     }
+
+    /// Verify that we can pass and receive an `Option<&RustType>`.
+    ///
+    /// We deinitialize the first reference and create a second to confirm that
+    /// deinitializing the reference does not deinitialize the Rust type.
+    func testSwiftCallRustWithOptionRefOpaqueRustType() throws {
+        let val = OptTestOpaqueRefRustType.new(123)
+        let opt_ref = val.field_ref()
+
+        let reflect = rust_reflect_option_ref_opaque_rust_type(opt_ref)
+        XCTAssertEqual(reflect!.field(), 123)
+        XCTAssertNil(rust_reflect_option_ref_opaque_rust_type(nil))
+        let reflect = nil
+        
+        let second_reference = rust_reflect_option_ref_opaque_rust_type(opt_ref)
+        XCTAssertEqual(second_reference!.field(), 123)
+    }
     
     func testSwiftCallRustWithOptionOpaqueRustCopyType() throws {
         let val = new_opaque_rust_copy_type(123)
--- a/crates/swift-bridge-ir/src/codegen/codegen_tests/option_codegen_tests.rs
+++ b/crates/swift-bridge-ir/src/codegen/codegen_tests/option_codegen_tests.rs
@@ -460,6 +460,124 @@ void* __swift_bridge__$some_function(void);
     }
 }
 
+/// Test code generation for Rust function that returns an Option<&OpaqueRustType>
+mod extern_rust_fn_return_option_ref_opaque_rust_type {
+    use super::*;
+
+    fn bridge_module_tokens() -> TokenStream {
+        quote! {
+            mod ffi {
+                extern "Rust" {
+                    type SomeType;
+                    fn some_function () -> Option<&SomeType>;
+                }
+            }
+        }
+    }
+
+    fn expected_rust_tokens() -> ExpectedRustTokens {
+        ExpectedRustTokens::Contains(quote! {
+            #[export_name = "__swift_bridge__$some_function"]
+            pub extern "C" fn __swift_bridge__some_function() -> *const super::SomeType {
+                if let Some(val) = super::some_function() {
+                    val as *const super::SomeType
+                } else {
+                    std::ptr::null()
+                }
+            }
+        })
+    }
+
+    fn expected_swift_code() -> ExpectedSwiftCode {
+        ExpectedSwiftCode::ContainsAfterTrim(
+            r#"
+func some_function() -> Optional<SomeTypeRef> {
+    { let val = __swift_bridge__$some_function(); if val != nil { return SomeTypeRef(ptr: val!) } else { return nil } }()
+}
+"#,
+        )
+    }
+
+    fn expected_c_header() -> ExpectedCHeader {
+        ExpectedCHeader::ContainsAfterTrim(
+            r#"
+void* __swift_bridge__$some_function(void);
+    "#,
+        )
+    }
+
+    #[test]
+    fn extern_rust_fn_return_option_opaque_rust_type() {
+        CodegenTest {
+            bridge_module: bridge_module_tokens().into(),
+            expected_rust_tokens: expected_rust_tokens(),
+            expected_swift_code: expected_swift_code(),
+            expected_c_header: expected_c_header(),
+        }
+        .test();
+    }
+}
+
+/// Test code generation for Rust function that returns an Option<&OpaqueRustType>
+mod extern_rust_fn_arg_option_ref_opaque_rust_type {
+    use super::*;
+
+    fn bridge_module_tokens() -> TokenStream {
+        quote! {
+            mod ffi {
+                extern "Rust" {
+                    type SomeType;
+                    fn some_function (arg: Option<&SomeType>);
+                }
+            }
+        }
+    }
+
+    fn expected_rust_tokens() -> ExpectedRustTokens {
+        ExpectedRustTokens::Contains(quote! {
+            #[export_name = "__swift_bridge__$some_function"]
+            pub extern "C" fn __swift_bridge__some_function(arg: *const super::SomeType) {
+                super::some_function(
+                    if arg.is_null() {
+                        None
+                    } else {
+                        Some( unsafe { & * arg })
+                    }
+                )
+            }
+        })
+    }
+
+    fn expected_swift_code() -> ExpectedSwiftCode {
+        ExpectedSwiftCode::ContainsAfterTrim(
+            r#"
+func some_function(_ arg: Optional<SomeTypeRef>) {
+    __swift_bridge__$some_function({ if let val = arg { return val.ptr } else { return nil } }())
+}
+"#,
+        )
+    }
+
+    fn expected_c_header() -> ExpectedCHeader {
+        ExpectedCHeader::ContainsAfterTrim(
+            r#"
+void __swift_bridge__$some_function(void* arg);
+    "#,
+        )
+    }
+
+    #[test]
+    fn extern_rust_fn_return_option_opaque_rust_type() {
+        CodegenTest {
+            bridge_module: bridge_module_tokens().into(),
+            expected_rust_tokens: expected_rust_tokens(),
+            expected_swift_code: expected_swift_code(),
+            expected_c_header: expected_c_header(),
+        }
+        .test();
+    }
+}
+
 /// Test code generation for Rust function that takes an Option<OpaqueRustType> argument.
 mod extern_rust_fn_with_option_opaque_rust_type_arg {
     use super::*;
--- a/crates/swift-integration-tests/src/option.rs
+++ b/crates/swift-integration-tests/src/option.rs
@@ -38,7 +38,6 @@ mod ffi {
         type OptTestOpaqueRefRustType;
 
         #[swift_bridge(init)]
-        #[swift_bridge(associated_to = OptTestOpaqueRustType)]
         fn new(field: u8) -> OptTestOpaqueRustType;
         fn field(self: &OptTestOpaqueRustType) -> u8;
 
--- a/crates/swift-integration-tests/src/option.rs
+++ b/crates/swift-integration-tests/src/option.rs
@@ -35,10 +35,16 @@ mod ffi {
 
     extern "Rust" {
         type OptTestOpaqueRustType;
+        type OptTestOpaqueRefRustType;
 
         #[swift_bridge(init)]
+        #[swift_bridge(associated_to = OptTestOpaqueRustType)]
         fn new(field: u8) -> OptTestOpaqueRustType;
-        fn field(&self) -> u8;
+        fn field(self: &OptTestOpaqueRustType) -> u8;
+
+        #[swift_bridge(associated_to = OptTestOpaqueRefRustType)]
+        fn new(field: u8) -> OptTestOpaqueRefRustType;
+        fn field_ref(self: &OptTestOpaqueRefRustType) -> Option<&OptTestOpaqueRustType>;
     }
 
     extern "Rust" {
@@ -84,6 +90,10 @@ mod ffi {
         fn rust_reflect_option_opaque_rust_type(
             arg: Option<OptTestOpaqueRustType>,
         ) -> Option<OptTestOpaqueRustType>;
+
+        fn rust_reflect_option_ref_opaque_rust_type(
+            arg: Option<&OptTestOpaqueRustType>,
+        ) -> Option<&OptTestOpaqueRustType>;
 
         fn rust_reflect_option_opaque_rust_copy_type(
             arg: Option<OptTestOpaqueRustCopyType>,
@@ -178,6 +188,22 @@ impl OptTestOpaqueRustType {
     }
 }
 
+pub struct OptTestOpaqueRefRustType {
+    field: Option<OptTestOpaqueRustType>,
+}
+
+impl OptTestOpaqueRefRustType {
+    fn new(field: u8) -> Self {
+        Self {
+            field: Some(OptTestOpaqueRustType::new(field)),
+        }
+    }
+
+    fn field_ref(&self) -> Option<&OptTestOpaqueRustType> {
+        self.field.as_ref()
+    }
+}
+
 #[derive(Copy, Clone)]
 pub struct OptTestOpaqueRustCopyType {
     #[allow(unused)]
@@ -240,6 +266,12 @@ fn rust_reflect_option_vector_rust_type(arg: Option<Vec<u16>>) -> Option<Vec<u16
 fn rust_reflect_option_opaque_rust_type(
     arg: Option<OptTestOpaqueRustType>,
 ) -> Option<OptTestOpaqueRustType> {
+    arg
+}
+
+fn rust_reflect_option_ref_opaque_rust_type(
+    arg: Option<&OptTestOpaqueRustType>,
+) -> Option<&OptTestOpaqueRustType> {
     arg
 }
 
--- a/crates/swift-bridge-ir/src/bridged_type/bridged_opaque_type.rs
+++ b/crates/swift-bridge-ir/src/bridged_type/bridged_opaque_type.rs
@@ -198,7 +198,12 @@ impl BridgeableType for OpaqueForeignType {
             let generics = self
                 .generics
                 .angle_bracketed_concrete_generics_tokens(types);
-            quote! { *mut super::#type_name #generics }
+
+            if self.reference {
+                quote! { *const super::#type_name #generics }
+            } else {
+                quote! { *mut super::#type_name #generics }
+            }
         }
     }
 
@@ -283,6 +288,16 @@ impl BridgeableType for OpaqueForeignType {
                     }
                 }
             }
+        } else if self.reference {
+            let ty = &self.ty;
+
+            quote! {
+                if let Some(val) = #expression {
+                    val as *const super::#ty
+                } else {
+                    std::ptr::null()
+                }
+            }
         } else {
             quote! {
                 if let Some(val) = #expression {
@@ -312,15 +327,23 @@ impl BridgeableType for OpaqueForeignType {
                     TypePosition::FnArg(func_host_lang, _)
                     | TypePosition::FnReturn(func_host_lang) => {
                         if func_host_lang.is_rust() {
-                            format!(
-                                "{{{}.isOwned = false; return {}.ptr;}}()",
-                                expression, expression
-                            )
+                            if self.reference {
+                                format!("{{return {}.ptr;}}()", expression)
+                            } else {
+                                format!(
+                                    "{{{}.isOwned = false; return {}.ptr;}}()",
+                                    expression, expression
+                                )
+                            }
                         } else {
-                            format!(
-                                "{{{}.isOwned = false; return {}.ptr;}}()",
-                                expression, expression
-                            )
+                            if self.reference {
+                                format!("{{return {}.ptr;}}()", expression)
+                            } else {
+                                format!(
+                                    "{{{}.isOwned = false; return {}.ptr;}}()",
+                                    expression, expression
+                                )
+                            }
                         }
                     }
                     TypePosition::SharedStructField => {
@@ -373,6 +396,11 @@ impl BridgeableType for OpaqueForeignType {
                         option_ffi_repr = option_ffi_repr,
                         ffi_repr = ffi_repr
                     )
+        } else if self.reference {
+            format!(
+                "{{ if let val = {expression} {{ return val.ptr }} else {{ return nil }} }}()",
+                expression = expression,
+            )
         } else {
             format!("{{ if let val = {expression} {{ val.isOwned = false; return val.ptr }} else {{ return nil }} }}()", expression = expression,)
         }
@@ -428,6 +456,14 @@ impl BridgeableType for OpaqueForeignType {
                     Some(unsafe{ #expression.val.assume_init() }.into_rust_repr())
                 } else {
                     None
+                }
+            }
+        } else if self.reference {
+            quote! {
+                if #expression.is_null() {
+                    None
+                } else {
+                    Some(unsafe {& * #expression} )
                 }
             }
         } else {
@@ -638,7 +674,11 @@ impl BridgeableType for OpaqueForeignType {
 
 impl OpaqueForeignType {
     pub fn swift_name(&self) -> String {
-        format!("{}", self.ty)
+        if self.reference {
+            format!("{}Ref", self.ty)
+        } else {
+            format!("{}", self.ty)
+        }
     }
 
     /// The name of the type used to pass a `#[swift_bridge(Copy(...))]` type over FFI


crates/swift-bridge-ir/src/codegen/codegen_tests/option_codegen_tests.rs
```diff
@@ -460,6 +460,64 @@ void* __swift_bridge__$some_function(void);
...
+                    std::ptr::null()
+                }
+            }
+        })
+    }
+
+    fn expected_swift_code() -> ExpectedSwiftCode {
+        ExpectedSwiftCode::ContainsAfterTrim(
+            r#"
+func some_function() -> Optional<SomeType> {
```
chinedufn, line 494: This should be `Optional<SomeTypeRef>`

crates/swift-bridge-ir/src/codegen/codegen_tests/option_codegen_tests.rs
```diff
@@ -460,6 +460,64 @@ void* __swift_bridge__$some_function(void);
...
+    { let val = __swift_bridge__$some_function(); if val != nil { return SomeType(ptr: val!) } else { return nil } }()
```
chinedufn, line 495: `SomeType` -> `SomeTypeRef`

SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/OptionTests.swift
```diff
@@ -97,6 +97,14 @@ class OptionTests: XCTestCase {
         
         XCTAssertNil(rust_reflect_option_opaque_rust_type(nil))
     }
+
+    func testSwiftCallRustWithOptionRefOpaqueRustType() throws {
+        let val = OptTestOpaqueRustTypeRef(123)
+        let reflect = rust_reflect_option_ref_opaque_rust_type(val)
```
chinedufn, line 103: I think that this might be UB at the moment, but should be easy to fix:

Let's add an extra call to `rust_reflect_option_ref_opaque_rust_type` so that we can be confident that we are returning a reference.

By calling it twice we can gain a bit more confidence that we aren't trying to free the reference.

So something along the lines of:

```swift
/// Verify that we can pass and receive an `Option<&RustType>`.
///
/// We deinitialize the first reference and create a second to confirm that
/// deinitializing the reference does not deinitialize the Rust type.
func testSwiftCallRustWithOptionRefOpaqueRustType() throws {
    let val = OptTestOpaqueRustTypeRef(123)

    let reflect = rust_reflect_option_ref_opaque_rust_type(val)
    XCTAssertEqual(reflect!.field(), 123)
    XCTAssertNil(rust_reflect_option_ref_opaque_rust_type(nil))
    let reflect = nil;

    let second_reference = rust_reflect_option_ref_opaque_rust_type(val);
    XCTAssertEqual(second_reference!.field(), 123)
}
```

---

From what I gather this test should fail. I'll comment on how to make it pass below.

chinedufn (commented):
Looks great. Thanks a lot for taking the time to put this together. 

Looks like you found the `if self.reference { quote! { *const super::#type_name #generics } }` bit. Awesome.

---

Just two changes please:

---

Avoid dropping the underlying Rust type when the Swift side drops the reference to the Rust type.
This is explained in the PR comments.

---

Update your PR body to show an example of a bridge module that did not work before this commit but works as of this commit.

Example: https://github.com/chinedufn/swift-bridge/pull/203#issue-1630545202

````
This commit adds support for Option<PrimitiveType> in `extern "Swift"`
functions.

For example, the following is now possible:

```rust
mod ffi {
    extern "Swift" {
        fn my_func(arg: Option<u8>) -> Option<bool>;
    }
}
```
````

---

After that this looks great and can land. Thank you for putting this together.

File: crates/swift-integration-tests/src/option.rs [hunk 4]
```diff
@@ -178,6 +191,20 @@ impl OptTestOpaqueRustType {
     }
 }
 
+pub struct OptTestOpaqueRustTypeRef {
+    field: u8,
+}
+
+impl OptTestOpaqueRustTypeRef {
+    fn new(field: u8) -> &'static Self {
+        &Self { field, }
```
chinedufn, line 200: This won't compile since we're trying to return a reference to a temporary value.

Can do something like:

```rust
pub struct OptTestOpaqueRustTypeRef {
    field: u8,
}

impl OptTestOpaqueRustTypeRef {
    fn new_some(field: u8) -> Option<Self> {
        Some(Self { field })
    }

    fn field(&self) -> u8 {
        self.field
    }
}

fn rust_convert_option_ty_to_option_ref_ty(val: Option<OptTestOpaqueRustTypeRef>) -> Option<&OptTestOpaqueRustTypeRef> {
    val.as_ref()
}
```

Then on the Swift side can call `OptTestOpaqueRustTypeRef.new_some` and then `rust_convert_option_ty_to_option_ref_ty` to get the `Option<&OptTestOpaqueRustTypeRef>` that you need.

---

Here's how to call an associated method such as `new_some` https://github.com/chinedufn/swift-bridge/blob/master/book/src/bridge-module/functions/README.md?plain=1#L146-L189

crates/swift-integration-tests/src/option.rs
```diff
@@ -41,6 +41,15 @@ mod ffi {
         fn field(&self) -> u8;
     }
 
+    extern "Rust" {
+        type OptTestOpaqueRustTypeRef;
+
+        #[swift_bridge(init)]
```
chinedufn, line 47: Can remove this `#[swift_bridge(init)]` as per the feedback below.

chinedufn (commented):
Left some notes on how to pass CI.

Also need to run `cargo fmt`.

crates/swift-bridge-ir/src/codegen/codegen_tests/option_codegen_tests.rs
```diff
@@ -460,6 +460,64 @@ void* __swift_bridge__$some_function(void);
     }
 }
 
+/// Test code generation for Rust function that returns an Option<&OpaqueRustType>
```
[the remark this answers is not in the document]
PrismaPhonic: Hahahaha, thank you. This is my first foray into FFI land so I appreciate it. 

I think I am missing what is needed here. Another `bridge_module_tokens()` method?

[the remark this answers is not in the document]
chinedufn: Gotcha:

1. Copy the module `extern_rust_fn_return_option_ref_opaque_rust_type` to a new module below it `extern_rust_fn_arg_option_ref_opaque_rust_type`

2. Replace the bridge module in `extern_rust_fn_arg_option_ref_opaque_rust_type` with the one in the comment above.

3. `extern_rust_fn_arg_option_ref_opaque_rust_type` will pass

4. Tweak the `expected_rust_tokens` to make it pass

crates/swift-bridge-ir/src/codegen/codegen_tests/option_codegen_tests.rs
PrismaPhonic: Done

SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/OptionTests.swift
PrismaPhonic: Updated so we check it twice

crates/swift-integration-tests/src/option.rs
PrismaPhonic: Done

crates/swift-integration-tests/src/option.rs
chinedufn: Ah whoops.

This is one solution:

```rust
fn rust_convert_ref_ty_to_option_ref_ty(val: &OptTestOpaqueRustTypeRef) -> Option<&OptTestOpaqueRustTypeRef> {
    Some(val)
}
``

Comment (PrismaPhonic) [reply crates/swift-integration-tests/src/option.rs]: Maybe something like this?

```rust
pub struct OptTestOpaqueRustTypeRef {
    field: Option<OptTestOpaqueRustType>,
}

impl OptTestOpaqueRustTypeRef {
    fn new(field: u8) -> Self {
        Some(Self { field: Some(OptTestOpaqueRustType::new(field)) })
    }

    fn field(&self) -> Option<OptTestOpaqueRustType> {
        self.field
    }

    fn field_ref(&self) -> &Option<OptTestOpaqueRustType> {
        &self.field
    }
}

fn rust_convert_ref_option_ty_to_option_ref_ty(
    val: &Option<OptTestOpaqueRustType>,
) -> Option<&OptTestOpaqueRustType> {
    val.as_ref()
}
```

chinedufn: Only challenge is that I'm not sure if `&Option<T>` is supported yet. SO the `val: &Option<OptTestOpaqueRustType>` might not work.

I like the general direction though. What about:

```rust
pub struct OptTestOpaqueRustTypeRef {
    field: Option<OptTestOpaqueRustType>,
}

impl OptTestOpaqueRustTypeRef {
    fn new(field: u8) -> Self {
        Some(Self { field: Some(OptTestOpaqueRustType::new(field)) })
    }

    fn field_ref(&self) -> Option<&OptTestOpaqueRustType> {
        self.field.as_ref()
    }
}
```

Then the test can call `.field_ref()` and then reflect the `Option<&OptTestOpaqueRustType>`.

PrismaPhonic: Oh right - that should work! Because then the parent object still owns the inner type. I'll try that :) thanks!

chinedufn: Team work made the dream work!

PrismaPhonic:
I've addressed all concerns listed here but seem to still hit this compiler error:

```rust
error[E0308]: mismatched types
   --> crates/swift-integration-tests/src/option.rs:95:13
    |
3   | #[swift_bridge::bridge]
    | ----------------------- arguments to this function are incorrect
...
95  |             arg: Option<&OptTestOpaqueRustType>,
    |             ^^^ types differ in mutability
    |
    = note: expected raw pointer `*mut _`
               found raw pointer `*const OptTestOpaqueRustType`
note: associated function defined here
   --> /home/pmfarr/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/boxed.rs:950:19
    |
950 |     pub unsafe fn from_raw(raw: *mut T) -> Self {
```

crates/swift-bridge-ir/src/codegen/codegen_tests/option_codegen_tests.rs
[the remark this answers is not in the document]
PrismaPhonic: Alright - did that to the best of my abilities and pushed up.

crates/swift-integration-tests/src/option.rs
PrismaPhonic: Alright, got that pushed up - hitting another compiler error (that I hit before) listed below in a general comment

chinedufn:
> Do you know if there's way to have the compiler show the error at the expanded code?

I think there's a `rustc` flag for it but I've never used it https://stackoverflow.com/a/68843349

> I think that we are generating a function that takes in a *const (which it should) but then tries to call Box::from_raw() on it which wants a *mut - but I'm not sure where to fix that

Yup. Can do another `if self.reference` in here:

https://github.com/chinedufn/swift-bridge/blob/4fbd30f81085dc366bda3c0303eac66345de3ed9/crates/swift-bridge-ir/src/bridged_type/bridged_opaque_type.rs#L438

Something like:

```rust
if self.reference {
    Some(unsafe {& * #expression})
} else {
    Some(unsafe { * Box::from_raw(#expression) } )
}
```

crates/swift-bridge-ir/src/codegen/codegen_tests/option_codegen_tests.rs
```diff
@@ -460,6 +460,122 @@ void* __swift_bridge__$some_function(void);
...
+mod extern_rust_fn_arg_option_ref_opaque_rust_type {
+    use super::*;
+
+    fn bridge_module_tokens() -> TokenStream {
+        quote! {
+            mod ffi {
+                extern "Rust" {
+                    type SomeType;
+                    fn some_function (arg: Option<&SomeType>) -> Option<&SomeType>;
```
chinedufn, line 530: I think this comment got swallowed since I was commenting on orphaned commits, so let me try here in the PR's "files changed" page.

Can remove the `-> Option<&SomeType>` here.

I made a mistake when I said to add it. The return case is already tested in the other test that you added so we don't need it here.

crates/swift-bridge-ir/src/codegen/codegen_tests/option_codegen_tests.rs
```diff
@@ -460,6 +460,122 @@ void* __swift_bridge__$some_function(void);
...
+            expected_rust_tokens: expected_rust_tokens(),
+            expected_swift_code: expected_swift_code(),
+            expected_c_header: expected_c_header(),
+        }
+        .test();
+    }
+}
+
+/// Test code generation for Rust function that returns an Option<&OpaqueRustType>
+mod extern_rust_fn_arg_option_ref_opaque_rust_type {
```
chinedufn, line 522: Just FYI here's a good test to take inspiration from: https://github.com/chinedufn/swift-bridge/blob/fdca9fcc2080a60d30de958fc4ada2516691581e/crates/swift-bridge-ir/src/codegen/codegen_tests/option_codegen_tests.rs#L463-L523

I'll comment below with the changes though, so no need to study that one too carefully.

crates/swift-bridge-ir/src/codegen/codegen_tests/option_codegen_tests.rs
```diff
@@ -460,6 +460,122 @@ void* __swift_bridge__$some_function(void);
...
+    fn expected_rust_tokens() -> ExpectedRustTokens {
+        ExpectedRustTokens::Contains(quote! {
+            #[export_name = "__swift_bridge__$some_function"]
+            pub extern "C" fn __swift_bridge__some_function() -> *const super::SomeType {
+                if let Some(val) = super::some_function() {
+                    val as *const super::SomeType
+                } else {
+                    std::ptr::null()
+                }
+            }
```
chinedufn, line 545: This can be:

```rust
#[export_name = "__swift_bridge__$some_function"]
pub extern "C" fn __swift_bridge__some_function(
    arg: *const super::SomeType
) {
    super::some_function(
        if arg.is_null() {
            None
        } else {
            Some( unsafe { & * arg } )
        }
    )
}
```

crates/swift-bridge-ir/src/codegen/codegen_tests/option_codegen_tests.rs
```diff
@@ -460,6 +460,122 @@ void* __swift_bridge__$some_function(void);
...
+        })
+    }
+
+    fn expected_swift_code() -> ExpectedSwiftCode {
+        ExpectedSwiftCode::ContainsAfterTrim(
+            r#"
+func some_function(arg: Optional<SomeTypeRef>) -> Optional<SomeTypeRef> {
+    { let val = __swift_bridge__$some_function(); if val != nil { return SomeTypeRef(ptr: val!) } else { return nil } }()
+}
```
chinedufn, line 554: This one is similar to: https://github.com/chinedufn/swift-bridge/blob/607e30579ac4de1fab5ff66838253cf3d07ddfbb/crates/swift-bridge-ir/src/codegen/codegen_tests/extern_rust_function_opaque_rust_type_argument_codegen_tests.rs#L91-L99

So it can be:

```swift
func some_function(_ arg: Optional<SomeTypeRef>) {
    __swift_bridge__$some_function({ if let val = arg { return val.ptr } else { return nil } }())
}
```

crates/swift-bridge-ir/src/codegen/codegen_tests/option_codegen_tests.rs
PrismaPhonic: So remove the entire module then? or...

chinedufn (commented):
I think this should do the trick

crates/swift-bridge-ir/src/codegen/codegen_tests/option_codegen_tests.rs
chinedufn: Keep the module but remove the `-> Option<&SomeType>` from this line 530 right here

So ` fn some_function (arg: Option<&SomeType>) -> Option<&SomeType>;`

becomes ` fn some_function (arg: Option<&SomeType>);

PrismaPhonic:
> > Do you know if there's way to have the compiler show the error at the expanded code?
> 
> I think there's a `rustc` flag for it but I've never used it https://stackoverflow.com/a/68843349
> 
> > I think that we are generating a function that takes in a *const (which it should) but then tries to call Box::from_raw() on it which wants a *mut - but I'm not sure where to fix that
> 
> Yup. Can do another `if self.reference` in here:
> 
> https://github.com/chinedufn/swift-bridge/blob/4fbd30f81085dc366bda3c0303eac66345de3ed9/crates/swift-bridge-ir/src/bridged_type/bridged_opaque_type.rs#L438
> 
> Something like:
> 
> ```rust
> if self.reference {
>     Some(unsafe {& * #expression})
> } else {
>     Some(unsafe { * Box::from_raw(#expression) } )
> }
> ```

Done, annnnnd now I'm hitting way more compiler errors :( 

```
error[E0424]: expected value, found module `self`
  --> crates/swift-integration-tests/src/option.rs:3:1
   |
3  | #[swift_bridge::bridge]
   | ^^^^^^^^^^^^^^^^^^^^^^^ `self` value is a keyword only available in methods with a `self` parameter
...
90 |         fn rust_reflect_option_opaque_rust_type(
   |            ------------------------------------ this function can't have a `self` parameter
   |
   = note: this error originates in the attribute macro `swift_bridge::bridge` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Somehow now we are generating functions that don't have self to have self. I'm soooooo confused

chinedufn:
> Done, annnnnd now I'm hitting way more compiler errors :(

Ah gotcha. Just need the `self` to be outside of the `quote!` :

```rust
if self.reference {
    quote! {
        if #expression.is_null() {
            None
        } else {
            Some(unsafe {& * #expression})
        }
    }
} else {
    quote! {
        if #expression.is_null() {
            None
        } else {
            Some(unsafe { * Box::from_raw(#expression) } )
        }
    }
}
```

crates/swift-bridge-ir/src/codegen/codegen_tests/option_codegen_tests.rs
PrismaPhonic: Seems that generated is this:

```swift
func some_function(_ arg: Optional<SomeTypeRef>) {
    __swift_bridge__$some_function({ if let val = arg { val.isOwned = false; return val.ptr } else { return nil } }())
}

```

Just checking that the assignment to setting isOwned to false is correct?

PrismaPhonic: It does by the way look like codegen on the previous test (that you told me to change to `return SomeTypeRef>` is actually `return SomeType` - I'm guessing this is a mistake and something I need to correct in translation to swift somewhere. Do you know where that would be?

chinedufn: `isOwned` is available on owned `SomeType` https://github.com/chinedufn/swift-bridge/blob/dd5bef56af28db4f1a1244d86115def6abede931/crates/swift-bridge-ir/src/codegen/codegen_tests/opaque_rust_type_codegen_tests.rs#L36-L48

but not on `SomeTypeRef` https://github.com/chinedufn/swift-bridge/blob/dd5bef56af28db4f1a1244d86115def6abede931/crates/swift-bridge-ir/src/codegen/codegen_tests/opaque_rust_type_codegen_tests.rs#L54-L60

---

So, we just need what should hopefully be one last `if self.reference` here https://github.com/chinedufn/swift-bridge/blob/4fbd30f81085dc366bda3c0303eac66345de3ed9/crates/swift-bridge-ir/src/bridged_type/bridged_opaque_type.rs#L312-L325

PrismaPhonic: Oh wait - that type has access to know if it's a reference - in that case I can just modify `swift_name` to check `if self.reference` and append the suffix accordingly

chinedufn: Hahaha look what I was about to comment with before I saw these last two comments:

https://github.com/chinedufn/swift-bridge/blob/4fbd30f81085dc366bda3c0303eac66345de3ed9/crates/swift-bridge-ir/src/bridged_type/bridged_opaque_type.rs#L639-L642

```rust
pub fn swift_name(&self) -> String {
	if self.reference {
		format!("{}", self.ty)
	} else {
		format!("{}Ref", self.ty)
	}
}
```

Your FFI-fu has grown strong my friend.

chinedufn: I think that as you start handling more cases the two languages start to diverge but based on what we currently have tested they're always the same.

As in, you end up starting to add certain if statements based on the language.

So what happened here was probably something like:

1. if rust do X, if swift `todo!()`
2. one day we needed to add support for Swift so we wrote a test case
3. then we changed the logic to -> if rust do X, if swift do X
4. FUTURE -> if rust do X or Y depending on Z, if swift do X or Q depending on W

So, this is just in the middle of that transition to step 4.

I suppose you could collapse then if you wanted, but I think it's better as is so that when someone someday needs to modify one language they can clearly see where to do it (I haven't had to do this in a while so I forget what can cause the divergence).

If you disagree and think we should collapse back down until that day comes, feel free to collapse it.

PrismaPhonic: One more spot I needed to check reference, but all working now!

In `convert_option_swift_expression_to_ffi_type`

PrismaPhonic:
@chinedufn addressed all issues - all tests are passing for me locally (using the command in the README)

crates/swift-bridge-ir/src/bridged_type/.bridged_opaque_type.rs.rustfmt
```diff
@@ -0,0 +1,764 @@
+use crate::bridged_type::{
```
chinedufn, line 1: Need to delete this file

SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/OptionTests.swift
```diff
@@ -97,6 +97,19 @@ class OptionTests: XCTestCase {
         
         XCTAssertNil(rust_reflect_option_opaque_rust_type(nil))
     }
+
+    func testSwiftCallRustWithOptionRefOpaqueRustType() throws {
```
chinedufn, line 101: Mind plopping this on here:

```rust
/// Verify that we can pass and receive an `Option<&RustType>`.
///
/// We deinitialize the first reference and create a second to confirm that
/// deinitializing the reference does not deinitialize the Rust type.
```

We'll eventually document every test in this way, so I want to enforce it for new ones.

The idea is that a new contributor can read the test docs and have a good sense of whether or not the test relates to what they're trying to do.

I was going to just add it myself but then I saw the `.rs.rustfmt` file so while commenting on that I figured I'd ask for this as well.

PrismaPhonic:
Huh, I wonder why integration tests failed in the pipeline but that one test in particular succeeds locally for me

crates/swift-integration-tests/src/option.rs
```diff
@@ -35,10 +35,16 @@ mod ffi {
 
     extern "Rust" {
         type OptTestOpaqueRustType;
+        type OptTestOpaqueRustTypeRef;
```
chinedufn, line 38: Need to rename this. the `type OptTestOpaqueRustType` above generates `class OptTestOpaqueRustTypeRef` on the Swift side, and then this also generates a class with the same name.

image

Can call it `OptTestOpaqueRefRustType` instead

chinedufn (approved):
Just need to delete a file and rename a type and then looks good to me.

Thanks for putting this together!``

crates/swift-integration-tests/src/option.rs
PrismaPhonic: ohhhhh right - that makes sense. Doh! magical strings hahaha

chinedufn: Here's more information as to why this error is happening. 

No need to read this link to solve the problem.
Renaming the type (I gave an example under the screenshot in the previous comment) will do.

https://github.com/chinedufn/swift-bridge/blob/dd5bef56af28db4f1a1244d86115def6abede931/book/src/bridge-module/opaque-types/README.md?plain=1#L52-L133

PrismaPhonic: Done and pushed up

PrismaPhonic: Hmm, actually, now I'm getting this failure locally:

```
Testing failed:
	Missing argument label 'ptr:' in call
	Cannot convert value of type 'Int' to expected argument type 'UnsafeMutableRawPointer'
	'nil' requires a contextual type
	Command SwiftCompile failed with a nonzero exit code
	```

Comment (PrismaPhonic) [reply crates/swift-integration-tests/src/option.rs]: I'm taking a wild guess here as I don't know swift that well but I'm guessing it's related to my logic here:

```swift
    func testSwiftCallRustWithOptionRefOpaqueRustType() throws {
        let val = OptTestOpaqueRefRustType.new(123)
        let opt_ref = val.field_ref()

        let reflect = rust_reflect_option_ref_opaque_rust_type(opt_ref)
        XCTAssertEqual(reflect!.field(), 123)
        XCTAssertNil(rust_reflect_option_ref_opaque_rust_type(nil))
        let reflect = nil
        
        let second_reference = rust_reflect_option_ref_opaque_rust_type(opt_ref)
        XCTAssertEqual(second_reference!.field(), 123)
    }
```

I wonder if this is the problem:

```swift
        XCTAssertEqual(reflect!.field(), 123)
```

Is it saying that the result of field() is a unsafe mutable raw pointer and it's trying to coerce the int of 123 into that but failing for the equality comparison?

In this case I call `val.field_ref()` which should get us the inner Option<&Type> and then I thought `reflect!.field()` goes inside that option and calls the `field()` method on the inner type which should then get us an int out?

SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/OptionTests.swift
PrismaPhonic: Good idea, done!

crates/swift-bridge-ir/src/bridged_type/.bridged_opaque_type.rs.rustfmt
PrismaPhonic: oh shoot! Sorry about that, thought I avoided committing that one. Done

SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/OptionTests.swift
```diff
@@ -97,6 +97,23 @@ class OptionTests: XCTestCase {
...
+    /// We deinitialize the first reference and create a second to confirm that
+    /// deinitializing the reference does not deinitialize the Rust type.
+    func testSwiftCallRustWithOptionRefOpaqueRustType() throws {
+        let val = OptTestOpaqueRefRustType.new(123)
+        let opt_ref = val.field_ref()
+
+        let reflect = rust_reflect_option_ref_opaque_rust_type(opt_ref)
+        XCTAssertEqual(reflect!.field(), 123)
+        XCTAssertNil(rust_reflect_option_ref_opaque_rust_type(nil))
+        let reflect = nil
```
chinedufn, line 112:
```swift
        var reflect = rust_reflect_option_ref_opaque_rust_type(opt_ref)
        XCTAssertEqual(reflect!.field(), 123)
        XCTAssertNil(rust_reflect_option_ref_opaque_rust_type(nil))
        reflect = nil
```

this way we can overwrite `reflect` and free the reference.

chinedufn:
Mind pushing new commits instead of force pushing? This way I can see what has changed. I'll squash them all at the end.

crates/swift-integration-tests/src/option.rs
```diff
@@ -35,10 +35,16 @@ mod ffi {
 
     extern "Rust" {
         type OptTestOpaqueRustType;
+        type OptTestOpaqueRefRustType;
 
         #[swift_bridge(init)]
+        #[swift_bridge(associated_to = OptTestOpaqueRustType)]
```
chinedufn, line 41: Need to remove this `associated_to`. I'll open an issue about emitting a compile time error t `#[swift_bridge(init)]` and `#[swift_bridge(associated_to = OptTestOpaqueRustType)]` from both being used.

Removing `associated_to` should fix the problem.

PrismaPhonic:
> Mind pushing new commits instead of force pushing? This way I can see what has changed. I'll squash them all at the end.

Can do - Github in particular has this problem unfortunately. In other platforms you can see a diff between rebases

crates/swift-integration-tests/src/option.rs
PrismaPhonic: Done - and all passing locally for me now :)

SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/OptionTests.swift
PrismaPhonic: Done

PrismaPhonic:
Addressed all comments - tests are passing for me locally now

crates/swift-integration-tests/src/option.rs
chinedufn: > I'll open an issue

Nevermind the real problem is that if there are two `#[swift_bridge(..)]` attributes the first one gets ignored.

I'll still track it. Thanks.

chinedufn:
Thanks for reporting this issue and thanks for putting together a fix. Appreciate you pushing through the unexpected compile time error after compile time error after compile time error.

Cheers.

PR #274, opened 2024-05-10
Add parse-bridges CLI subcommand

integer-overflown:
## What's this?

A new `parse-bridges` subcommand.
Also, an update to the documentation in the example showing this feature's preview.

This is handy for usage in higher-level build systems.
For example, in my case, I build both Rust and Swift with CMake, and having this CLI command implemented would allow generating the glue code from CMake as well, as a dependency step before building Rust/Swift targets (via `add_custom_target` API).

## Notes

One caveat of the current implementation is that one would have to duplicate the crate name in the invocation line (the first argument).

This is fine for cases like mine, where this would be taken from the build system anyway, but it may not be handy for other cases.

The package name can be automatically deduced if one's running in the correct directory (the package root).
In this case, we can parse the `cargo read-manifest` output and get the name from there.

This would require a new dependency, though (`serde_json`), so I decided not to do this just yet, but if this sounds okay to you, I'll go ahead and implement this as well.

## Examples

Single file:
```shell
swift-bridge-cli parse-bridges --crate-name my-awesome-crate -f src/lib.rs -o generated
```

Multiple files:
```shell
swift-bridge-cli parse-bridges --crate-name my-superb-crate \
-f src/lib.rs \
-f src/some_other_file.rs \
-o generated
```

### DIFF
--- a/book/src/building/swiftc-and-cargo/README.md
+++ b/book/src/building/swiftc-and-cargo/README.md
@@ -140,7 +140,7 @@ when you have `Cargo` linking in the Swift code.
 ```rust
 // build.rs
 fn main() {
-    println!("cargo:rustc-link-lib=static=swiftc_link_rust");
+    println!("cargo:rustc-link-lib=static=my_swift");
     println!("cargo:rustc-link-search=./");
 }
 ```
--- a/book/src/building/swiftc-and-cargo/README.md
+++ b/book/src/building/swiftc-and-cargo/README.md
@@ -70,7 +70,7 @@ func run() {
 ```
 
 ```sh
-# briding-header.h
+# bridging-header.h
 
 #ifndef BridgingHeader_h
 #define BridgingHeader_h
@@ -134,24 +134,12 @@ chmod +x build-swiftc-links-rust.sh
 
 ## Rust links to a Swift native library
 
-Unlike the when we had `swiftc` in the Rust code, you do not need to set the `crate-type`
+Unlike when we had `swiftc` linking in the Rust code, you do not need to set the `crate-type`
 when you have `Cargo` linking in the Swift code.
 
 ```rust
 // build.rs
-use std::path::PathBuf;
-
 fn main() {
-    let out_dir = PathBuf::from("./generated");
-
-    let bridges = vec!["src/lib.rs"];
-    for path in &bridges {
-        println!("cargo:rerun-if-changed={}", path);
-    }
-
-    swift_bridge_build::parse_bridges(bridges)
-        .write_all_concatenated(out_dir, env!("CARGO_PKG_NAME"));
-
     println!("cargo:rustc-link-lib=static=swiftc_link_rust");
     println!("cargo:rustc-link-search=./");
 }
@@ -163,9 +151,10 @@ fn main() {
 #!/bin/bash
 set -e
 
-# The swift-bridge CLI does not exist yet. Open an issue if you need to use
-# this approach and I'll happily whip up the CLI.
-swift-bridge -f src/lib.rs > generated
+# Generate the bridging headers.
+# Multiple files can be supplied with "-f" flag, e.g. "-f file1 -f file2".
+# Substitute <package name> for the crate name in your Cargo.toml file.
+swift-bridge-cli parse-bridges --crate-name <package name> -f src/lib.rs -o generated
 
 swiftc -emit-library -static -module-name my_swift -import-objc-header bridging-header.h \
   lib.swift ./generated/swift-and-rust/swift-and-rust.swift
--- a/crates/swift-bridge-cli/src/clap_app.rs
+++ b/crates/swift-bridge-cli/src/clap_app.rs
@@ -1,4 +1,4 @@
-use clap::{Arg, Command};
+use clap::{Arg, ArgAction, Command};
 
 /// The CLI application
 pub fn cli() -> Command<'static> {
@@ -7,6 +7,7 @@ pub fn cli() -> Command<'static> {
         .version(env!("CARGO_PKG_VERSION"))
         .subcommand_required(true)
         .subcommand(create_package_command())
+        .subcommand(create_bridges_command())
 }
 
 /// The command for creating a Swift Package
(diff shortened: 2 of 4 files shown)


crates/swift-bridge-cli/src/clap_exec.rs
```diff
@@ -34,3 +37,21 @@ fn handle_create_package(matches: &ArgMatches) {
...
+
+    let crate_name = matches
+        .get_one::<String>("crate-name")
+        .unwrap(); // required
+    let source_files: Vec<&String> = matches.get_many("source-files")
+        .unwrap()
+        .collect(); // required
+    let output = matches.get_one::<String>("output")
+        .map(Path::new)
+        .unwrap_or_else(|| Path::new(DEFAULT_OUTPUT_DIR));
```
chinedufn, line 53: Let's remove the default and require an explicit output directory.

Then we can consider introducing a default in the future if we decide that needing to specify the directory yourself is too cumbersome.

Reasons:

1. One fewer thing to figure out during this review. Adding a default would be a non-breaking change.

2. Fewer defaults helps teach the user about how things work / what is configurable

3. It is likely that users will write this command once and then re-use it, so it's okay if it's a little longer

book/src/building/swiftc-and-cargo/README.md
```diff
@@ -163,9 +163,7 @@ fn main() {
 #!/bin/bash
 set -e
 
-# The swift-bridge CLI does not exist yet. Open an issue if you need to use
-# this approach and I'll happily whip up the CLI.
-swift-bridge -f src/lib.rs > generated
+swift-bridge-cli parse-bridges <package name> src/lib.rs
```
chinedufn, line 166: Let's use `--file / -f` for each source file:

Let's use `--crate-name` for the crate name:

Reasons:

- allow arguments to be specified in any order
- more clear which argument is which. Right now `<package name>` and `<file name>` are right next to each other without any flags, making it a little harder to understand what each argument does.

crates/swift-bridge-cli/src/clap_app.rs
```diff
@@ -101,3 +102,34 @@ fn create_package_command() -> Command<'static> {
...
+                .multiple_values(true)
+                .required(true)
+        )
+        .arg(
+            Arg::new("output")
+                .long("output")
+                .short('o')
+                .takes_value(true)
+                .value_name("PATH")
+                .help("Output destination folder; subfolders are created automatically")
```
chinedufn, line 133: No need to mention that subfolders are created automatically.
This information isn't useful to the user.

chinedufn (commented):
Thanks for putting this together. Left some minor feedback.

---

> One caveat of the current implementation is that one would have to duplicate the crate name in the invocation line (the first argument).

I'm not understanding what you mean here.

> but if this sounds okay to you

Currently no known use cases where people would be repeatedly typing out this command by hand.

So, let's hold off on simplifying it and prefer explicit/obvious behavior for now.

---

Please update your PR body with an example of how the command looks.

When writing release notes we go to the PRs to grab examples.

We'll also sometimes link someone to an old similar PR, so it's nice to be able to look at the PR and immediately see what it's accomplishing.

integer-overflown:
Thanks for reviewing.

> I'm not understanding what you mean here.

Inferring crate name from the manifest if a user is located in the package root.

> Currently no known use cases where people would be repeatedly typing out this command by hand.

> So, let's hold off on simplifying it and prefer explicit/obvious behavior for now.

I agree; this would be less error-prone and more explicit.
As you also mentioned, this command will likely be written once, and then integrated into a script or build-system rules.

> Please update your PR body with an example of how the command looks.

Good point; done, see the examples section.

crates/swift-bridge-cli/src/clap_exec.rs
integer-overflown: Good point, done.

I've force-pushed the changes.

book/src/building/swiftc-and-cargo/README.md
integer-overflown: Done; updates the docs as well

crates/swift-bridge-cli/src/clap_app.rs
integer-overflown: I found this helpful :) 

But I might be in the minority here, so I removed this.

book/src/building/swiftc-and-cargo/README.md
```diff
@@ -134,24 +134,12 @@ chmod +x build-swiftc-links-rust.sh
...
-
-    let bridges = vec!["src/lib.rs"];
-    for path in &bridges {
-        println!("cargo:rerun-if-changed={}", path);
-    }
-
-    swift_bridge_build::parse_bridges(bridges)
-        .write_all_concatenated(out_dir, env!("CARGO_PKG_NAME"));
-
     println!("cargo:rustc-link-lib=static=swiftc_link_rust");
```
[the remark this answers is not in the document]
chinedufn: Yup nice catch. Can make this change then I'll land this PR.

book/src/building/swiftc-and-cargo/README.md
```diff
@@ -134,24 +134,12 @@ chmod +x build-swiftc-links-rust.sh
...
 fn main() {
-    let out_dir = PathBuf::from("./generated");
-
-    let bridges = vec!["src/lib.rs"];
-    for path in &bridges {
-        println!("cargo:rerun-if-changed={}", path);
-    }
-
-    swift_bridge_build::parse_bridges(bridges)
-        .write_all_concatenated(out_dir, env!("CARGO_PKG_NAME"));
```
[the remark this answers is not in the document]
chinedufn: Looks like the file now includes an example of generating the headers via API and via CLI, so this sounds good.

chinedufn:
Thanks for putting this together. Can land once you:

- fix the `my_swift.a` that you noticed
- get tests passing (looks like you need to run `cargo fmt`)

integer-overflown:
@chinedufn done, pushed the changes :)

integer-overflown:
@chinedufn I've pushed another commit that fixes the tests.

At least, on my side, `cargo test --all` passes successfully.

I'm not sure how my changes broke the UI tests.
There was probably some update on the CI runner side, which made the compiler more verbose and caused the test to fail.

I've put an explainer in the commit message.

Hopefully, it passes now 😅

NiwakaDev:
> I'm not sure how my changes broke the UI tests.

@integer-overflown 

You need to address https://github.com/chinedufn/swift-bridge/pull/276#discussion_r1637913279.

chinedufn:
Congrats on landing your first contribution. Thank you for pushing this to completion. Good stuff.

PR #282, opened 2024-07-19
Implement Error for RustStringRef

jnbooth:
With the latest versions of Swift, using `Result<_, String>` causes two errors:
- `'RustString' is not convertible to 'any Error'`
- `Thrown expression type 'RustString' does not conform to 'Error`
 
The solution to both errors is for `RustStringRef` to declare conformance to the `Error` protocol.

See: [Swift Documentation - Error](https://developer.apple.com/documentation/swift/error)

### DIFF
--- a/crates/swift-bridge-build/src/generate_core/rust_string.swift
+++ b/crates/swift-bridge-build/src/generate_core/rust_string.swift
@@ -48,11 +48,6 @@ extension RustStringRef {
     }
 }
 extension RustStringRef: Error {}
-extension RustStringRef: LocalizedError {
-  public var errorDescription: String? {
-    return self.as_str().toString()
-  }
-}
 extension RustString: Vectorizable {
     public static func vecOfSelfNew() -> UnsafeMutableRawPointer {
         __swift_bridge__$Vec_RustString$new()
--- a/crates/swift-bridge-build/src/generate_core/rust_string.swift
+++ b/crates/swift-bridge-build/src/generate_core/rust_string.swift
@@ -47,6 +47,12 @@ extension RustStringRef {
         __swift_bridge__$RustString$trim(ptr)
     }
 }
+extension RustStringRef: Error {}
+extension RustStringRef: LocalizedError {
+  public var errorDescription: String? {
+    return self.as_str().toString()
+  }
+}
 extension RustString: Vectorizable {
     public static func vecOfSelfNew() -> UnsafeMutableRawPointer {
         __swift_bridge__$Vec_RustString$new()


crates/swift-bridge-build/src/generate_core/rust_string.swift
```diff
@@ -47,6 +47,12 @@ extension RustStringRef {
         __swift_bridge__$RustString$trim(ptr)
     }
 }
+extension RustStringRef: Error {}
+extension RustStringRef: LocalizedError {
+  public var errorDescription: String? {
+    return self.as_str().toString()
+  }
+}
```
chinedufn, line 55: The PR described why we need an `Error` implementation.

It does not mention `LocalizedError`.
Why do we need this? Please remove or justify the addition.

After that this looks good and can land.

chinedufn (commented):
Thanks for making this contribution.

One question about why you introduced the `LocalizedError` implementation.

crates/swift-bridge-build/src/generate_core/rust_string.swift
jnbooth: Oh, good point! I should have mentioned that in the PR. `LocalizedError` is a specialization of `Error`. All errors have a [`.localizedDescription`](https://developer.apple.com/documentation/swift/error/localizeddescription) virtual field that describes the error in text. For example, here's a very naive error handler:

```swift
func handleError(_ error: Error) {
  print(error.localizedDescription)
}
```

If a type only extends `Error`, its `.localizedDescription` field is nothing. By implementing `LocalizedError`, the type chooses what `.localizedDescription` will be. Here, the field becomes the string itself. So `print(error.localizedDescription)` will directly print out the error message.

As a separate task, we could have `RustStringRef` implement [`CustomStringConvertible`](https://developer.apple.com/documentation/swift/customstringconvertible), and then `print(error)` would print out the error message directly. But that would be a bigger architectural decision, and it would have both pros and cons. Plus, that's out of scope of this PR.

jnbooth: Sounds good. I removed the extension.

chinedufn:
Just needs an integration test.

# Potential approach

Bridge a function here that returns `-> Result<(), String>`, somewhere in here https://github.com/chinedufn/swift-bridge/blob/de059024732fa2f8135a9fe1232b727d96794af6/crates/swift-integration-tests/src/result.rs#L3-L81

Then add a test case that calls that function somewhere inside here https://github.com/chinedufn/swift-bridge/blob/de059024732fa2f8135a9fe1232b727d96794af6/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunnerTests/ResultTests.swift#L10

On the test case document that (I'm paraphrasing) around the time of `Swift x.y` returning `-> Result<(), String>` started giving errors like `Thrown expression type 'RustString' does not conform to 'Error'`. And that we solved that by implementing Swift's `Error` protocol for `RustStringRef`. And that confirms the the `Error` protocol is implemented.

Then confirm that if you comment out the `extension RustStringRef: Error {}` line then tests fail.

tmolitor-stud-tu:
@jnbooth Do you still work on this?

chinedufn:
Implemented in https://github.com/chinedufn/swift-bridge/pull/296

PR #287, opened 2024-08-12
Support throwing initializers

NiwakaDev:
This PR implements throwing initializers. See: https://docs.swift.org/swift-book/documentation/the-swift-programming-language/errorhandling/.

Here's an example of using this feature:

```rust
// Rust
#[swift_bridge::bridge]
mod ffi {
    enum ResultTransparentEnum {
        NamedField { data: i32 },
        UnnamedFields(u8, String),
        NoFields,
    }
    extern "Rust" {
        type ThrowingInitializer;
        #[swift_bridge(init)]
        fn new(succeed: bool) -> Result<ThrowingInitializer, ResultTransparentEnum>;
        fn val(&self) -> i32;
    }
}
```

```swift
// Swift
do {
    let throwingInitializer = try ThrowingInitializer(false)
} catch let error as ResultTransparentEnum {
    //...
} catch {
    //...
}
```

### DIFF
--- a/crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
+++ b/crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
@@ -1,4 +1,5 @@
 use crate::bridged_type::{BridgeableType, BridgedType, CFfiStruct, TypePosition};
+use crate::parse::HostLang;
 use crate::{TypeDeclarations, SWIFT_BRIDGE_PREFIX};
 use proc_macro2::{Span, TokenStream};
 use quote::{format_ident, quote, quote_spanned};
@@ -207,6 +208,7 @@ impl BuiltInResult {
                 }
                 "__private__ResultPtrAndPtr".to_string()
             }
+            TypePosition::ThrowingInit(_) => todo!(),
         }
     }
 
@@ -253,6 +255,18 @@ impl BuiltInResult {
                 ),
                 TypePosition::SharedStructField => todo!(),
                 TypePosition::SwiftCallsRustAsyncOnCompleteReturnTy => todo!(),
+                TypePosition::ThrowingInit(lang) => {
+                    match lang {
+                        HostLang::Rust => format!(
+                            "let val = {expression}; switch val.tag {{ case {c_ok_name}: self.init(ptr: val.payload.ok) case {c_err_name}: throw {err_swift_type} default: fatalError() }}",
+                        expression = expression,
+                        c_ok_name = c_ok_name,
+                        c_err_name = c_err_name,
+                        err_swift_type = err_swift_type
+                    ),
+                        HostLang::Swift => todo!(),
+                    }
+                }
             };
         }
 
--- a/crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
+++ b/crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
@@ -258,10 +258,9 @@ impl BuiltInResult {
                 TypePosition::ThrowingInit(lang) => {
                     match lang {
                         HostLang::Rust => format!(
-                            "let val = {expression}; switch val.tag {{ case {c_ok_name}: self.init(ptr: val.payload.ok) case {c_err_name}: throw {err_swift_type} default: fatalError() }}",
+                            "let val = {expression}; if val.tag == {c_ok_name} {{ self.init(ptr: val.payload.ok) }} else {{ throw {err_swift_type} }}",
                         expression = expression,
                         c_ok_name = c_ok_name,
-                        c_err_name = c_err_name,
                         err_swift_type = err_swift_type
                     ),
                         HostLang::Swift => todo!(),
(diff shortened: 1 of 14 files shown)


crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
```diff
@@ -253,6 +255,18 @@ impl BuiltInResult {
                 ),
                 TypePosition::SharedStructField => todo!(),
                 TypePosition::SwiftCallsRustAsyncOnCompleteReturnTy => todo!(),
+                TypePosition::ThrowingInit(lang) => {
+                    match lang {
+                        HostLang::Rust => format!(
+                            "let val = {expression}; switch val.tag {{ case {c_ok_name}: self.init(ptr: val.payload.ok) case {c_err_name}: throw {err_swift_type} default: fatalError() }}",
```
NiwakaDev, line 261: I guess that it's not suitable because the `convert_ffi_value_to_swift_value` method returns a expression.

chinedufn: I'm not sure what you're communicating?

crates/swift-bridge-ir/src/codegen/codegen_tests/opaque_rust_type_codegen_tests.rs
```diff
@@ -527,3 +527,82 @@ void* __swift_bridge__$Foo$new(void);
         .test();
     }
 }
+
+/// Verify that we can use a Swift class with a throwing init.
```
chinedufn, line 531: ~use a Swift class~ "create a Swift class"

crates/swift-bridge-ir/src/codegen/codegen_tests/opaque_rust_type_codegen_tests.rs
```diff
@@ -527,3 +527,82 @@ void* __swift_bridge__$Foo$new(void);
...
+
+    deinit {
+        if isOwned {
+            __swift_bridge__$Foo$_free(ptr)
+        }
+    }
+}
+extension Foo {
+    public convenience init() throws {
+        let val = __swift_bridge__$Foo$new(); switch val.tag { case __swift_bridge__$ResultFooAndSomeErrEnum$ResultOk: self.init(ptr: val.payload.ok) case __swift_bridge__$ResultFooAndSomeErrEnum$ResultErr: throw val.payload.err.intoSwiftRepr() default: fatalError() }
```
chinedufn, line 584: Let's use an `if` statement instead of a `switch` so that we don't need this `default: fatalError()`

The `default: fatalError` makes it seem like this can fail, even though it can't.

crates/swift-bridge-ir/src/parse/parse_extern_mod/function_attributes.rs
```diff
@@ -295,7 +296,10 @@ mod tests {
 
         let func = &module.functions[0];
         assert!(func.is_swift_initializer);
-        assert!(func.is_swift_failable_initializer);
+        matches!(
+            func.swift_failable_initializer,
+            Some(FailableInitializerType::Option)
+        );
```
chinedufn, line 302: Need a test for `FailableInitializerType::Throwing`

chinedufn (commented):
Let's document `#[swift_bridge(init)]` in the `Function Attributes` section in the book

https://github.com/chinedufn/swift-bridge/blob/bdbb028dea65148136851ff24bb8a025c5c66557/book/src/bridge-module/functions/README.md?plain=1#L66

Can add a `#[swift_bridge(init)]` section.

Can include examples of:

- regular `init`
- throwing `init`
- option `init`

crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
chinedufn: I think you're asking whether it's okay to return two statements here instead of a single expression?

This should be fine since the `ThrowingInit` code here is only ever used inside of an initializer and isn't ever embedded inside of another expression.

chinedufn:
Great work on this. Looks good. Left some minor feedback.

crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs
NiwakaDev: The other return values of `convert_ffi_value_to_swift_value` return closures, but this case doesn't	return a closure. This means that returns a simple statement, not expression.

For examples:
https://github.com/chinedufn/swift-bridge/blob/495611ba39bc6b08838408751b92f6329b98f321/crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs#L286

https://github.com/chinedufn/swift-bridge/blob/495611ba39bc6b08838408751b92f6329b98f321/crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs#L247

The above examples return closures.

crates/swift-bridge-ir/src/codegen/codegen_tests/opaque_rust_type_codegen_tests.rs
NiwakaDev: Done. However, I don't address the others yet. 

For example:
https://github.com/chinedufn/swift-bridge/blob/495611ba39bc6b08838408751b92f6329b98f321/crates/swift-bridge-ir/src/bridged_type/bridgeable_result.rs#L247

I'll address this in the next PR, or we can open it as good first issue.

NiwakaDev:
@chinedufn 
Done.

chinedufn:
Looks great thanks.

PR #290, opened 2024-08-26
fix: wrongly detect as a virtual manifest

setoelkahfi:
The `Cargo.toml` file needs a `[package]` section; otherwise, the `[lib]` section is not allowed. See [virtual manifest](https://doc.rust-lang.org/cargo/reference/workspaces.html#virtual-workspace).


chinedufn:
Pleasure to accept your first contribution.

Issue #107, opened 2022-08-11
`rust-binary-calls-swift-package` example doesn't work on a Linux environment


yoching:
Hello.
I'm trying to build a CLI application that is mainly written in Rust and calls some Swift functions, and I'm developing it on a Linux environment.
I tried to run `rust-binary-calls-swift-package` example, but it didn't run with following errors.
I'm using WSL2 with Ubuntu and swift-5.6.2-RELEASE-ubuntu20.04.

Do you know how I can make it work? Or, are Linux environments not supported?
Thank you in advance!

```sh
$ cargo run -p rust-binary-calls-swift-package
   Compiling proc-macro2 v1.0.42
   Compiling unicode-ident v1.0.2
   Compiling quote v1.0.20
   Compiling syn v1.0.98
   Compiling libc v0.2.126
   Compiling fastrand v1.8.0
   Compiling remove_dir_all v0.5.3
   Compiling cfg-if v1.0.0
   Compiling tempfile v3.3.0
   Compiling swift-bridge-ir v0.1.35 (/home/yoshikuni/GitHub/Rust/swift-bridge/crates/swift-bridge-ir)
   Compiling swift-bridge-build v0.1.35 (/home/yoshikuni/GitHub/Rust/swift-bridge/crates/swift-bridge-build)
   Compiling swift-bridge-macro v0.1.35 (/home/yoshikuni/GitHub/Rust/swift-bridge/crates/swift-bridge-macro)
   Compiling rust-binary-calls-swift-package v0.1.0 (/home/yoshikuni/GitHub/Rust/swift-bridge/examples/rust-binary-calls-swift-package)
   Compiling swift-bridge v0.1.35 (/home/yoshikuni/GitHub/Rust/swift-bridge)
error: linking with `cc` failed: exit status: 1
  |
[...]

PR #297, opened 2024-10-23
Make `rust-binary-calls-swift-package` example support Linux

terraputix:
Closes https://github.com/chinedufn/swift-bridge/issues/107.

### DIFF
--- a/examples/rust-binary-calls-swift-package/build.rs
+++ b/examples/rust-binary-calls-swift-package/build.rs
@@ -51,6 +51,17 @@ fn main() {
         // Without this there will be a lot of missing symbols when linking to the Swift library.
         let swift_lib_path = get_swift_lib_path().unwrap();
         println!("cargo:rustc-link-search={}", swift_lib_path);
+
+        // These linker flags are needed if `cargo test` is executed, `cargo run` works without them.
+        // Still, the order is important and they need to be placed before swiftCore.
+        println!("cargo:rustc-link-lib=FoundationEssentials");
+        println!("cargo:rustc-link-lib=_FoundationCShims");
+        println!("cargo:rustc-link-lib=swift_StringProcessing");
+        println!("cargo:rustc-link-lib=swift_RegexParser");
+        println!("cargo:rustc-link-lib=swiftGlibc");
+        println!("cargo:rustc-link-lib=_FoundationCollections");
+
+        // These swift libraries are needed in any case.
         println!("cargo:rustc-link-lib=swiftCore");
         println!("cargo:rustc-link-lib=stdc++");
         println!("cargo:rustc-link-lib=swiftSwiftOnoneSupport");
--- a/examples/rust-binary-calls-swift-package/build.rs
+++ b/examples/rust-binary-calls-swift-package/build.rs
@@ -48,54 +48,36 @@ fn main() {
     // This fix is for Linux only
     #[cfg(target_os = "linux")]
     {
-        // Without this there will be a lot of missing symbols when linking to the Swift library.
-        let swift_lib_path = get_swift_lib_path().unwrap();
+        // We need to tell cargo which additional libraries to link to!
+        //
+        // This is required because swift build -Xswiftc -static-stdlib works for executables,
+        // but not for libraries (yet). Thus, when trying to link against the produced .a file,
+        // not all symbols can be resolved. The undefined symbols can easily be found by running
+        // `nm -u .build/debug/libswift-library.a`, usually things such as `swift_retain`
+        // and `swift_release` will be missing.
+        // Cargo will give you an error message like this if symbols are missing:
+        // note: /usr/bin/ld: .build/debug/libswift-library.a(swift_library.swift.o): in function `$ss27_finalizeUninitializedArrayySayxGABnlF':
+        //       <compiler-generated>:(.text+0x17): undefined reference to `$sSaMa'
+        // or:                                      undefined reference to `swift_release'
+        //
+        // Thus, we need to explicitly link against the Swift libraries which are required.
+        // Unfortunately, the required linker flags depend on the Swift version and the used modules,
+        // so they might be different for your project.
+        let swift_lib_path = std::env::var("SWIFT_LIBRARY_PATH")
+            .unwrap_or_else(|_| "/usr/lib/swift/linux".to_string());
+
+        if !std::path::Path::new(&swift_lib_path).exists() {
+            panic!("Swift library path not found at /usr/lib/swift/linux and SWIFT_LIBRARY_PATH environment variable not set");
+        }
+
         println!("cargo:rustc-link-search={}", swift_lib_path);
 
-        // These linker flags are needed if `cargo test` is executed, `cargo run` works without them.
-        // Still, the order is important and they need to be placed before swiftCore.
-        println!("cargo:rustc-link-lib=FoundationEssentials");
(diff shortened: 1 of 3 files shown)


examples/rust-binary-calls-swift-package/build.rs
```diff
@@ -17,48 +17,90 @@ fn main() {
...
 }
 
 fn compile_swift() {
     let swift_package_dir = manifest_dir().join("swift-library");
 
     let mut cmd = Command::new("swift");
 
-    cmd.current_dir(swift_package_dir)
-        .arg("build")
-        .args(&["-Xswiftc", "-static"])
```
terraputix, line 52: I removed this because the Package.swift already defines this as a static library.

examples/rust-binary-calls-swift-package/build.rs
```diff
@@ -17,48 +17,90 @@ fn main() {
...
-    // For example:
-    // ld: warning: Could not find or use auto-linked library 'swiftCompatibility51'
-    // ld: warning: Could not find or use auto-linked library 'swiftCompatibility50'
-    // ld: warning: Could not find or use auto-linked library 'swiftCompatibilityDynamicReplacements'
-    // ld: warning: Could not find or use auto-linked library 'swiftCompatibilityConcurrency'
-    let xcode_path = if let Ok(output) = std::process::Command::new("xcode-select")
-        .arg("--print-path")
-        .output()
+    // This fix is for macOS only
+    #[cfg(target_os = "macos")]
```
terraputix, line 21: I am not sure if this is also required for other Apple platforms?

chinedufn: Not sure, but the line `"cargo:rustc-link-search={}/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/macosx/` suggests that it's `macos` only.

examples/rust-binary-calls-swift-package/build.rs
```diff
@@ -17,48 +17,101 @@ fn main() {
...
+
+    // This fix is for Linux only
+    #[cfg(target_os = "linux")]
+    {
+        // Without this there will be a lot of missing symbols when linking to the Swift library.
+        let swift_lib_path = get_swift_lib_path().unwrap();
+        println!("cargo:rustc-link-search={}", swift_lib_path);
+
+        // These linker flags are needed if `cargo test` is executed, `cargo run` works without them.
+        // Still, the order is important and they need to be placed before swiftCore.
```
chinedufn, line 56: Can you make this comment explain why the order is important?

We want users reading this example to understand exactly what is going on.

examples/rust-binary-calls-swift-package/build.rs
```diff
@@ -17,48 +17,101 @@ fn main() {
...
+        println!("cargo:rustc-link-lib=_FoundationCollections");
+
+        // These swift libraries are needed in any case.
+        println!("cargo:rustc-link-lib=swiftCore");
+        println!("cargo:rustc-link-lib=stdc++");
+        println!("cargo:rustc-link-lib=swiftSwiftOnoneSupport");
+    }
+}
+
+fn get_swift_lib_path() -> Result<String, String> {
```
chinedufn, line 71: Is there a default path for the Swift library?

If so, instead of spawning a `which` command can we just use the default path, and if it doesn't exist we can panic and tell the user to set a `SWIFT_LIBRARY_PATH=...` environment variable that points to "explanation of what to point to here".

This way:

- we avoid spawning an extra command
- the example more clearly explains its dependencies

Right now this is a bit magical, whereas an example should be more explicit and help the user understand exactly what's happening.

examples/rust-binary-calls-swift-package/build.rs
```diff
@@ -17,48 +17,101 @@ fn main() {
...
+    }
+
+    // This fix is for Linux only
+    #[cfg(target_os = "linux")]
+    {
+        // Without this there will be a lot of missing symbols when linking to the Swift library.
+        let swift_lib_path = get_swift_lib_path().unwrap();
+        println!("cargo:rustc-link-search={}", swift_lib_path);
+
+        // These linker flags are needed if `cargo test` is executed, `cargo run` works without them.
```
chinedufn, line 55: Can you make the comment explain why they are only needed in `cargo test`, but not `cargo run`?

examples/rust-binary-calls-swift-package/build.rs
```diff
@@ -17,48 +17,101 @@ fn main() {
...
+            "cargo:rustc-link-search={}/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/macosx/",
+            &xcode_path
+        );
+        println!("cargo:rustc-link-search={}", "/usr/lib/swift");
+    }
+
+    // This fix is for Linux only
+    #[cfg(target_os = "linux")]
+    {
+        // Without this there will be a lot of missing symbols when linking to the Swift library.
```
chinedufn, line 51: Can you give an example of one or a few of the errors?

This helps someone who has run into that error find this example when they search online.

chinedufn (commented):
Looks good. Left some minor feedback.

There have been multiple issues/PRs lately about Linux support.
There's at least some level of demand for `swift-bridge` to support Swift+Linux.
This PR would be helpful.

Hopefully you're still able to complete it after such a long delay. I can review your fixes quickly.

Thanks for working on this.

examples/rust-binary-calls-swift-package/build.rs
terraputix: I adapted the code to use `/usr/lib/swift/linux` as a default path, if not otherwise configured by a user via the `SWIFT_LIBRARY_PATH` environment variable.

examples/rust-binary-calls-swift-package/build.rs
terraputix: I cannot reproduce this anymore with swift 6.0.3. and rust 1.83.0 on fedora

examples/rust-binary-calls-swift-package/build.rs
terraputix: I added an example.

examples/rust-binary-calls-swift-package/build.rs
terraputix: I cannot reproduce this anymore with swift 6.0.3. and rust 1.83.0 on fedora.

examples/rust-binary-calls-swift-package/build.rs
terraputix: Hm, maybe it is problematic in cross-compilation scenarios?

terraputix:
I had another look at this PR and got it working on a newer Swift version (most likely I still used 5.9 or 5.10 when submitting this PR).
This fix gets the job done on my machine, but I fear it might not be the most generic fix. 

I was also quickly trying to get this built correctly using the [static-linux-sdk](https://www.swift.org/documentation/articles/static-linux-getting-started.html), but I had no luck with that.

I guess the real issue is that Swift Package Manager does not correctly support static linking for libraries, but only for executables, as this recent issue suggests: https://github.com/swiftlang/swift-package-manager/issues/8198

chinedufn:
Got it. Well, before this PR a user couldn't run on Linux at all, so this is an improvement.

If people have problems with these instructions they can open issues and we can try to continue to make the example to work on more Swift versions.

Thanks for landing your first contribution. People will appreciate this example.

Issue #306, opened 2025-01-26
Suggestion: use server-side swift to run integration tests


colinmarc:
Hi, thanks for the great library!

Currently, running the integration test suite (which is fantastic) requires xcodebuild. It seems to me that it should be easy to use server-side swift for this, which would allow running the integration tests on different platforms (for example, linux CI runners).

chinedufn:
No problem, glad you're poking around.

---

Questions:

- How difficult would it be to port the test suite from Xcode -> using `swiftc` directly (or whatever "server-side" Swift is)?
  - Why is it "easy"? What work would be required?

- Would all integration tests be possible to run outside of Xcode, or would we still need Xcode for some tests?

- Are there any trade-offs to ditching Xcode for our integration test suite?

---

Yeah, enabling folks that don't have Apple hardware to run the test suite would be great.

colinmarc:
>     * How difficult would it be to port the test suite from Xcode -> using `swiftc` directly (or whatever "server-side" Swift is)?

I believe it would require no changes to the swift code, only to the test harness. I've never used XCTest without XCode, but googling around it seems to be possible.

>    * Why is it "easy"? What work would be required?

1. The rust-side code would have to be wrapped as a swift package instead of added directly to the harness xcproject. I believe there's already an example of this elsewhere in the repo.
2. The test runner itself would be a swift package rather than an xcode project.

>     * Would all integration tests be possible to run outside of Xcode, or would we still need Xcode for some tests?

I don't know the answer to this, but given that the tests seem to be mostly pure swift, I can't imagine many would depend on xcode. Are there any tests using platform stuff like SwiftUI?

>     * Are there any trade-offs to ditching Xcode for our integration test suite?

Some people like xcode, and they would (probably) have to drop to the command line to run the integration test suite.

chinedufn:
Got it, thanks.

I just did some quick research.

#### Some Relevant Links

- `Swift on Server` 
  - https://www.swift.org/documentation/server/
- `Swift Testing`
  - https://www.swift.org/documentation/server/guides/testing.html
  - https://github.com/swiftlang/swift-testing

---

Looks like anyone with the Swift CLI can run `swift test` to use `Swift Testing`, Swift's new test harness / tools.

---

Path forward:

- we create a new empty directory for our integration tests. We'll gradually migrate old tests to this new directory.
 
[...]

PR #307, opened 2025-01-26
Introduce a new SPM-based test harness

colinmarc:
Ref. #306

This is currently pretty hacky, and also doesn't actually compile on my machine because Swift 6 introduced a bunch of concurrency checks.


test-swift-rust-integration.sh
```diff
@@ -11,17 +7,22 @@ THIS_DIR=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
...
-# During the build process these will get overwritten with their real final contents.
-touch ./Generated/SwiftBridgeCore.{h,swift}
-mkdir -p ./Generated/swift-integration-tests
-touch ./Generated/swift-integration-tests/swift-integration-tests.{h,swift}
+# Compile the rust library used by the integration tests.
+cargo build --manifest-path crates/swift-integration-tests/Cargo.toml
+
+# Create a swift file with the generated swift code and an import for the
+# generated C code. This is necessary because SPM projects don't support
+# mixed-source - the C code has to be a separate SPM target.
```
colinmarc, line 15: This is something that should be addressed upstream before this goes any farther. Basically, unlike, xcode, SPM doesn't allow mixed-language modules - you can't put `.swift` and `.h` files in the same target. Right now, the generation code assumes that all the swift and C code will coexist in the same namespace.

colinmarc:
With #312 and #310, this now runs to completion! :tada: 

```
[33/33] Linking IntegrationTestsPackageTests.xctest
Build complete! (4.24s)
Test Suite 'All tests' started at 2025-01-27 14:53:36.569
Test Suite 'debug.xctest' started at 2025-01-27 14:53:36.570
Test Suite 'PrimitiveTests' started at 2025-01-27 14:53:36.570
Test Case 'PrimitiveTests.testRustCallsSwiftPrimitives' started at 2025-01-27 14:53:36.570
Test Case 'PrimitiveTests.testRustCallsSwiftPrimitives' passed (0.0 seconds)
Test Case 'PrimitiveTests.testSwiftCallsRustPrimitives' started at 2025-01-27 14:53:36.570
Test Case 'PrimitiveTests.testSwiftCallsRustPrimitives' passed (0.1 seconds)
Test Suite 'PrimitiveTests' passed at 2025-01-27 14:53:36.670
	 Executed 2 tests, with 0 failures (0 unexpected) in 0.101 (0.101) seconds
Test Suite 'debug.xctest' passed at 2025-01-27 14:53:36.670
	 Executed 2 tests, with 0 failures (0 unexpected) in 0.101 (0.101) seconds
Test Suite 'All tests' passed at 2025-01-27 14:53:36.670
	 Executed 2 tests, with 0 failures (0 unexpected) in 0.101 (0.101) seconds
◇ Test run started.
↳ Testing Library Version: 6.0.3 (18c42c19cac3faf)
✔ Test run with 0 tests passed after 0.001 seconds.
```

integration-tests/Tests/IntegrationTests/PrimitiveTests.swift
```diff
@@ -0,0 +1,28 @@
+import RustLib
```
colinmarc, line 1: This is just a reformat of the existing test, plus this `import RustLib` was added. If you'd like me to take a stab at porting it to the new [Swift Testing](https://developer.apple.com/xcode/swift-testing/) framework, let me know.

chinedufn:
Thanks for working on all of this stuff. I haven't looked yet.
My current plan is to to review this and your other issues either Thurs or Fri.

integration-tests/Tests/IntegrationTests/PrimitiveTests.swift
```diff
@@ -0,0 +1,28 @@
...
+import XCTest
+
+@testable import SharedLib
+
+/// Tests for generic types such as `type SomeType<u32>`
```
chinedufn, line 6: Awesome. Let's add a `README.md` file to this `integration-tests` dir.

It should include:

- instructions on how to run this `integration-tests` test suite
- a note that old tests are being gradually ported here, such that it becomes clearer to new contributors that their new tests should go in this dir and not in the old Xcode integration test dir
- anything else that you think would be useful for maintainers to know

integration-tests/Tests/IntegrationTests/PrimitiveTests.swift
```diff
@@ -0,0 +1,28 @@
...
+class PrimitiveTests: XCTestCase {
+    /// Run tests where Rust calls Swift functions that take primitive args.
+    func testRustCallsSwiftPrimitives() throws {
+        test_rust_calls_swift_primitives()
+    }
+
+    /// Run tests where Swift calls Rust functions that take primitive args.
+    func testSwiftCallsRustPrimitives() throws {
```
chinedufn, line 14: In the old test suite you could click on a test in Xcode to run just that test.

In this new test suite, let's leave a comment above each test with the command to run just that test.

````
/// ## To Run
/// ```
/// swift test .... args for running just this test here ...
/// ```
````


Additionally, let's also add a command above the `class PrimitiveTests` that can be used to run all the primitive tests.

The idea is for a contributor to be able to easily figure out how to run the tests for the part of the system that they are working on.

Currently, most contributors have more Rust experience than Swift experience, so more hand-holding on the Swift side is useful.

File: integration-tests/Package.swift [hunk 5]
```diff
@@ -0,0 +1,26 @@
+// swift-tools-version: 6.0
+
+import PackageDescription
+
+let linkerSettings: [LinkerSetting] = [
+    .linkedLibrary("swift_integration_tests"),
+    .unsafeFlags(["-L../target/debug/"])
```
chinedufn, line 7: Can you see if there is a way to make this conditionally use debug or release based on an environment variable?

This makes it easier for someone to look at this script and adapt it to their own application's needs.

We can default to debug.

integration-tests/Package.swift
```diff
@@ -0,0 +1,26 @@
...
+]
+
+let package = Package(
+    name: "IntegrationTests",
+    targets: [
+        // The compiled static rust library.
+        .systemLibrary(
+            name: "RustLib"),
+        // The generated Swift wrapper code for the Rust library, plus some
+        // Swift code used by the Rust library.
```
chinedufn, line 17: Let's be more explicit about exactly what this "generated Swift wrapper code" and "Swift code used by the Rust library is".

A new contributor should be able to read this and know where to look if they want more information.

So:

- clarify that the generated Swift wrapper code is the FFI bindings that `swift-bridge` generates so that the languages can call into each other
- clarify that "Swift code used by the Rust library" is the `swift-bridge` core code that is always generated for all projects, and mention the name of the file that this code gets created in.
  - `crates/swift-bridge-build/src/generate_core.rs`

This level of clarity will help maintainers that are navigating the test suite more quickly understand where to look next when they want to fix something.

integration-tests/Package.swift
```diff
@@ -0,0 +1,26 @@
+// swift-tools-version: 6.0
```
chinedufn, line 1: This file is very clean. Nice. Definitely an improvement over the Xcode integration tests pile of configuration files.

crates/swift-integration-tests/src/result.rs
```diff
@@ -10,8 +10,6 @@ mod ffi {
         ) -> Result<ResultTestOpaqueRustType, ResultTestOpaqueRustType>;
 
         fn rust_func_takes_result_string(arg: Result<String, String>);
-        fn rust_func_returns_result_string(ok: bool) -> Result<String, String>;
```
chinedufn, line 13: Why was this removed?

crates/swift-integration-tests/build.rs
```diff
@@ -1,8 +1,8 @@
 use std::path::PathBuf;
 
 fn main() {
-    let out_dir = "../../SwiftRustIntegrationTestRunner/Generated";
-    let out_dir = PathBuf::from(out_dir);
+    // TODO: we can use --artifact-dir with OUT_DIR when that stabilizes.
```
chinedufn, line 4: What do you mean by "with `OUT_DIR`"? Why would we use both?

Why would we want to use either of these things instead of the hard coded string that we're using now.

Let's make it clear what this TODO is trying to achieve. Why is this beneficial?

This is important because 5 years from now someone else that is working on this might not understand what your intentions were here.

We try to be very clear with our TODOs so that anyone can pick them up when they come across them.

---

Also, let's link to the tracking issue so that a maintainer can more easily check if the feature that we're waiting for it available. https://github.com/rust-lang/cargo/issues/6790

integration-tests/Tests/IntegrationTests/PrimitiveTests.swift
chinedufn, line 1: Let's add a `GitHub` actions job that runs these `integration-tests`.

chinedufn (commented):
Awesome work here.
Excited for it to be possible for Linux users to contribute to the project.

---

Left some mostly minor feedback and questions.

chinedufn:
Please update the README to include more information so that:

- when I rebase these commits I can use the PR body as the commit message
- contributors can more quickly look at a PR and understand 60%+ of what it is doing / why it exists / the motivations
- when writing release notes we can more quickly find the important details to communicate

I left notes about this here -> https://github.com/chinedufn/swift-bridge/pull/310#issuecomment-2625358930

chinedufn:
I left my review as if this PR was ready and wasn't a draft, but just a heads up that I recognize that this was submitted as a draft so if I may have commented on stuff that you were already going to change.

Figured that since the PR was small it was easier to make sure we were both on the same page vs. making assumptions about what was finished and what wasn't.

test-swift-rust-integration.sh
```diff
@@ -1,8 +1,4 @@
-#!/bin/bash
-
-# Integration tests between Swift and Rust
-
-set -e
```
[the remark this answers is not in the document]
colinmarc: I just moved it up to the shebang out of habit. Happy to restore it if you prefer this idiom.

integration-tests/.swift-format
```diff
@@ -0,0 +1,10 @@
+{
```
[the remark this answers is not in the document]
colinmarc: I would strongly suggest having this file somewhere, so that future PRs aren't fighting over indentation. The defaults are (imo) quite unpleasant, but if you want to use the defaults we can also just create a file with the defaults.

crates/swift-integration-tests/build.rs
colinmarc: I thought it was self-evident that hardcoding the output path is brittle (I stumbled over this). Letting `integration-tests.sh` dictate the output directory seems much better. I can add more to the TODO explicating that.

test-swift-rust-integration.sh
```diff
@@ -11,17 +7,19 @@ THIS_DIR=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
...
+# generated C code. This is necessary because SPM projects don't support
+# mixed-source - the C code has to be a separate SPM target.
+OUT_DIR="integration-tests/Sources/Generated"
+DST="integration-tests/Sources/SharedLib/Generated/SharedLib.swift"
+echo "import RustLib" > "$DST"
+cat "${OUT_DIR}/SwiftBridgeCore.swift" >> "$DST"
+cat "${OUT_DIR}/swift-integration-tests/swift-integration-tests.swift" >> "$DST"
+
+# Run the swift package tests. Note that we have to instruct swift where to look
+# for the rust static lib.
```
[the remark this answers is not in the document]
colinmarc: Oh, it was in this line of code, and then I moved it inside `Package.swift`.

Cargo.toml
```diff
@@ -33,8 +33,6 @@ members = [
   "crates/swift-bridge-macro",
 
   "crates/swift-integration-tests",
-  "SwiftRustIntegrationTestRunner/integration-test-create-swift-package",
-  "SwiftRustIntegrationTestRunner/swift-package-rust-library-fixture",
```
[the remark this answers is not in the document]
colinmarc: It's a bit awkward to have both the old and new integration test suite in the same PR, so my intention was to keep this as a draft until all current tests are migrated over (verified by running the tests and counting the assertions, or whatever). The reason is this:

 - The swift code depends on the `swift-integration-tests` crate
 - That crate depends on the swift code in `SwiftRustIntegrationTestRunner` (Result.swift, etc)
 - Therefore, that code has to be duplicated in both harnesses, or there has two be two different identical rust crates.

If you favor a more incremental approach, I would propose the following:

 - Leave the existing harness untouched
 - Create a blank "project" as the new test harness, with no rust code and no swift code. (Or maybe just the primitive tests as a starter.)
 - Incrementally add tests to it, adding both the needed rust code and needed swift code, in a series of PRs.

crates/swift-bridge-ir/src/codegen/codegen_tests/result_codegen_tests.rs
```diff
@@ -4,7 +4,7 @@ use super::{CodegenTest, ExpectedCHeader, ExpectedRustTokens, ExpectedSwiftCode}
 use proc_macro2::TokenStream;
 use quote::quote;
 
-/// Test code generation for Rust function that accepts a Result<T, E>
```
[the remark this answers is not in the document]
colinmarc: This is part of the revert in #310, which I rebased in so that it would compile.

crates/swift-bridge-ir/src/codegen/codegen_tests/result_codegen_tests.rs
```diff
@@ -64,72 +64,6 @@ void __swift_bridge__$some_function(struct __private__ResultPtrAndPtr arg);
     }
 }
 
-/// Test code generation for Rust function that returns a Result<T, E>
```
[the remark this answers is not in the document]
colinmarc: This is part of #310.

crates/swift-integration-tests/src/result.rs
colinmarc: Also part of #310.

integration-tests/Package.swift
colinmarc: Maybe, but in general this stems from the problem that creating a SPM package from a swift-bridge crate requires a lot of different bits of configuration to match up. I will explicate that in a new issue (also talking about the imports).

test-swift-rust-integration.sh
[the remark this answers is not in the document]
chinedufn: Oh I didn't notice/know about that. Ok moving it up is fine.

crates/swift-integration-tests/build.rs
chinedufn: Ah, I have a better sense now.

What I have in mind was that, since this path will likely change less than once every 3+ years, flexibility isn't much of a concern.

So, we'd be trading away the ability to do `cargo build` on this crate for the flexibility of overriding a flag that we never need to override.
Similarly, users that are using this script as inspiration for their own codebase won't benefit from the flexibility since they also probably always want to us the same out dir 100% of the time, for years in a row.

If you still disagree then I'm cool with proceeding with your plan here, just add a reason to the TODO since, at least for me, it isn't obvious that it's an improvement.

> I can add more to the TODO explicating that.

Great. That'd be enough for me. Thanks.

integration-tests/.swift-format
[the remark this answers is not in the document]
chinedufn: Ok so having this file is required for Swift's formatter to work? Got it.

Cool, I have no experience with Swift's formatter so let's go with the settings you've chosen.

Cargo.toml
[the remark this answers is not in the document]
chinedufn: > Therefore, that code has to be duplicated in both harnesses, or there has two be two different identical rust crates.

I'm not quite following why we'd need to duplicate anything vs. restructuring things such that we can move Rust+Swift code over from the old suite to the new suite, one batch at time.
i.e., `Old Swift + Old Rust (starts with everything) -> New Swift + New Rust (starts with nothing)`

Maybe that's what your incremental approach proposal is saying. I'm not quite sure if you're suggesting that there are downsides to the incremental approach.

---

Regardless, I'm cool with whichever option you prefer.
I'm just happy to see this work progressing and won't bother trying to weigh the pros and cons, so anything that best helps you land it is more than fine by me.

test-swift-rust-integration.sh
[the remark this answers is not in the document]
chinedufn: Got it. What I meant was "Make the comment mention where the instruction lives so that the reader knows where to jump to next if they're trying to step through how this all works".

test-swift-rust-integration.sh
colinmarc: #315. I think taking the approach I outlined there would address a lot of the issues here, for example the hardcoded output path in `build.rs`.

Cargo.toml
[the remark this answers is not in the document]
colinmarc: It's probably a bit slower, but it's also cleaner. Once #310 and #312 land, I will rebase and push a version that does that (and then mark the PR as not a draft).

integration-tests/Tests/IntegrationTests/PrimitiveTests.swift
colinmarc: Good idea.

integration-tests/Package.swift
colinmarc: (#315)

integration-tests/.swift-format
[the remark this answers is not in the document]
colinmarc: Sorry, to be clear, it isn't strictly necessary. I just recommend it so that future contributors won't reformat every file they touch in case they have different defaults configured.

Cargo.toml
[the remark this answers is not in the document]
chinedufn: I've replied to #310 and #312 with recommendations for how to land them.

Hoping to unblock this `Introduce a new SPM-based test harness` PR so that we can merge it and have a foundation that we can start to build upon (by gradually porting tests over).

debanjanbasu (approved):

PR #296, opened 2024-10-14
Support Rust returning `-> Result<_, String>`

brittlewis12:
Support Rust returning `-> Result<_, String>`, e.g.:

```rust
#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        fn do_fallible_work() -> Result<(), String>;
    }
}
```

---

Incorporate the changes from #282 to implement swift's `Error` protocol for ruststring, & extend them with the tests requested in https://github.com/chinedufn/swift-bridge/issues/295#issuecomment-2411049833

---

thanks for your work on this lovely crate!

Issue #269, opened 2024-04-15
Support `#[swift_bridge(Sendable)]` to implement `Rust Send+Sync` for `Swift Sendable` types


chinedufn:
Related: `Implementing Swift's Error and Sendable protocols on an opaque Rust type's generated Swift class` - https://github.com/chinedufn/swift-bridge/issues/150

---

Right now when generating the Rust representation of a Swift type we create a non-Send and non-Sync type.

```rust
#[swift_bridge::bridge]
mod ffi {
    extern "Swift" {
        type SomeSwiftType;
    }
}

// Generated code
// This DOES NOT implement Send or Sync
struct SomeSwiftType(*mut std::ffi::c_void);
```

If the user knows that `SomeSwiftType` is `Sendable` then they should be able to `Send + Sync` on the Rust side.

```rust
#[swift_bridge::bridge]
mod ffi {
    extern "Swift" {
        #[swift_bridge(Sendable)]
        type AnotherSwiftType;
    }
}

// Generated Rust code
struct AnotherSwiftType(std::ffi::c_void);
unsafe impl Send for AnotherSwiftType {}
unsafe impl Sync for AnotherSwiftType {}
```

Then it would generate Swift code like:

```swift
// This would appear once in `SwiftBridgeCore.swift`
func __assert_sendable<__SwiftType: Sendable>() {}

// This is auto-generated.
// It confirms that `AnotherSwiftType` does in fact
// implement Swift's `Sendable` protocol.
func check_AnotherSwiftType_is_Sendable() {
    __assert_sendable::<AnotherSwiftType>()
}
```

chinedufn:
Implemented https://github.com/chinedufn/swift-bridge/pull/317

Issue #155, opened 2023-02-03
Enforce ownership using Swift's consume operator


chinedufn:
Swift recently accepted a [proposal for a `consume` operator][consume].

The `consume` operator will let us declare that a function takes unique ownership of a type https://github.com/apple/swift-evolution/blob/main/proposals/0366-move-function.md?plain=1#L175-L184 .

Any attempts to use the type after it is consumed lead to a compile time error.

This new `consume` operator will make it possible for us to solve one of the memory safety issues described in the book https://github.com/chinedufn/swift-bridge/blob/master/book/src/safety/README.md?plain=1#L48-L73 .

After the `consume` operator lands in stable Swift we can use it on any methods that take an owned `self`.

So, the following bridge module:

```rust
#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        type Computer
        fn shutdown(self);

        fn eat(&self, other: Computer);
    }
}
```

would generate Swift code along the lines of:

```swift
// Generated Swift (approximation)

class Computer {
    // Notice the `__consuming`
    func __consuming shutdown() { /* */ }

    // Notice the `consume`.
    func eat(other: consume Computer) { /* */ }
}
```

I'm not sure when the `consume` operator will land or how we can track its progress.

[consume]: https://github.com/apple/swift-evolution/blob/main/proposals/0366-move-function.md

chinedufn:
Swift 5.9 introduces some new ownership features. I haven't thought through what we can do with all of them yet. Linking to some resources here:

- Looks like a fairly comprehensive overview of some of the new ownership features https://www.hackingwithswift.com/articles/258/whats-new-in-swift-5-9

- A section of a WWDC talk that briefly discusses some ownership features https://developer.apple.com/videos/play/wwdc2023/10164/?time=1405

chinedufn:
More resources:

- https://github.com/apple/swift-evolution/blob/main/proposals/0390-noncopyable-structs-and-enums.md
- https://github.com/apple/swift-evolution/blob/main/proposals/0366-move-function.md
- https://github.com/apple/swift-evolution/blob/main/proposals/0377-parameter-ownership-modifiers.md

chinedufn:
I think a good approach would be to introduce Swift compile-time enforced ownership support behind an attribute in the `swift-bridge = "0.1.*"`, then after things stabilize we can remove the attribute, make Swift ownership the default behavior and remove our old runtime enforced ownership.
[...]

Issue #150, opened 2023-01-29
Error and Sendable conformance


chinedufn:
## The Problem

In [#149] we implemented support for returning `-> Result<T, E>` where `E` is an opaque Rust type.

```rust
// Rust

#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        type TypeA;
        type TypeB;

        fn some_function() -> Result<TypeA, TypeB>;

        fn print_info(self: &TypeB);
    }
}
```

Swift errors need to conform to Swift's `Error` and `Sendable` protocols, but right now there is no way to
declare that an Opaque Rust type should conform to `Error` and `Sendable`.

Instead, you currently have to manually specify that a type conforms.

https://github.com/chinedufn/swift-bridge/blob/2ab37cbb949be4a466848129e7f17ecbb3fecf5e/SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunner/Result.swift#L26-L27

## Desired Future

If an `OpaqueRustType` implements Rust's `Send` and `Sync` traits then a user should be able to emit a `Sendable` protocol implementation.

```rust
#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        #[swift_bridge(Sendable)]
        type TypeA;
    }
}

// This implements `Send` and `Sync`
struct TypeA {
    field: Vec<u8>
}
```

On the Swift side this would generate:

```swift
// Swift

extension TypeA: @unchecked Sendable {}
extension TypeA: Error {}
```

### Sendable

We need to have a good understanding of Swift's `Sendable` protocol's requirements.

The `Resources` section below contains links with more information.

### Swift reference types

The `struct TypeA` in Rust becomes a `class TypeA` in Swift.

If this `class TypeA` implements `Sendable` then it can be sent across threads,
meaning that two different threads could access the mutable methods on the instance of
`class TypeA` at the same time.

Is it possible for us to prevent this? Could Swift's Actor's be of use here?

### Concurrency

In general we need to think through Swift's concurrency model and make sure that we are going
down a safe path.

Getting some design feedback from someone with a good amount of Swift experience would be helpful.

## Resources

- [Swift Sendable documentation](https://developer.apple.com/documentation/swift/sendable)

- [Sendable explained](https://www.avanderlee.com/swift/sendable-protocol-closures/)

[#149]: https://github.com/chinedufn/swift-bridge/pull/149

NiwakaDev:
[...]

PR #312, opened 2025-01-27
Add Sendable to all async return types in tests

colinmarc:
Without this fix I get errors like the following under Swift 6:

```
/home/colinmarc/dev/swift-bridge/integration-tests/Sources/SharedLib/Generated/SharedLib.swift:4159:26: error: sending 'rustFnRetVal' risks causing data races
4157 |     return try await withCheckedThrowingContinuation({ (continuation: CheckedContinuation<AsyncResultOkEnum, Error>) in
4158 |         let callback = { rustFnRetVal in
4159 |             continuation.resume(with: rustFnRetVal)
     |                          |- error: sending 'rustFnRetVal' risks causing data races
     |                          `- note: task-isolated 'rustFnRetVal' is passed as a 'sending' parameter; Uses in callee may race with later task-isolated uses
4160 |         }
4161 |
```

I avoided adding `Sendable` for `RustString` since that's being discussed in #296 and elsewhere. Instead, I commented it out and left a TODO.


SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunner/Result.swift
```diff
@@ -41,5 +43,7 @@ extension ResultTransparentStruct: Error {}
 extension SameEnum: @unchecked Sendable {}
 extension SameEnum: Error {}
 
+extension AsyncResultOkEnum: @unchecked Sendable {}
```
chinedufn, line 46: Hmm we'll want to move away from these `@unchecked Sendable {}`s in favor of enabling `swift-bridge` to emit `Sendable` extension.

In the meantime, you can make this `@unchecked Sendable` safe by replacing the inner `String` with a primitive such as `u32`.

https://github.com/chinedufn/swift-bridge/blob/636fa2774870c052a18f06d98ca5f6966c09bdd6/crates/swift-integration-tests/src/async_function.rs#L50-L54

Let's also leave a TODO in this file to ditch `@unchecked Sendable` in favor of proper `Sendable` support. Can link to #269

SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunner/Result.swift
```diff
@@ -6,23 +6,25 @@
...
         self.num = val
     }
-    
+
     func val() -> UInt32 {
         self.num
     }
 }
 
+extension AsyncRustFnReturnStruct: @unchecked Sendable {}
```
chinedufn, line 26: Currently safe since this struct only contains a primitive https://github.com/chinedufn/swift-bridge/blob/636fa2774870c052a18f06d98ca5f6966c09bdd6/crates/swift-integration-tests/src/async_function.rs#L8-L11

But we'll still eventually want to replace all of these `@unchecked Sendable` with proper `swift-bridge` support for emitting `Sendable` impls as described in #269

crates/swift-integration-tests/src/async_function.rs
```diff
@@ -13,7 +13,8 @@ mod ffi {
     extern "Rust" {
         async fn rust_async_return_null();
         async fn rust_async_reflect_u8(arg: u8) -> u8;
-        async fn rust_async_reflect_string(string: String) -> String;
+        // TODO: this is broken because RustString is not Sendable.
+        // async fn rust_async_reflect_string(string: String) -> String;
```
chinedufn, line 17: The `TODO` describes the problem but not what we should do about it.

Do we need to do research? Do we already know the solution? Is the problem unsolvable? etc.
Let's make it more clear here what the next step is on this TODO.

chinedufn (commented):
Minor feedback around documentation and changing a `String` to a `u32`, then we can land this

SwiftRustIntegrationTestRunner/SwiftRustIntegrationTestRunner/Result.swift
colinmarc: For future reference, I think you linked the wrong issue by accident - the relevant one is #150.

colinmarc:
I think I addressed your comments.

colinmarc:
Hm, I don't think `RustStr` should be made sendable until the work in #155 lands. I would merge this as-is and then follow up reenabling the `String`-based tests after it does.

chinedufn:
We should make it Sendable and, for now, put the onus on the user to maintain the safety rules outlined in the book.

If in the future we're able to guarantee safety at compile time, even better.

In the meantime, if users aren't following the book's safety rules then they're going to hit UB regardless of whether we implement `Sendable`.

---

Swift does not yet make it possible for a fully safe FFI boundary. We need to accept that and work with it, rather than wishing that the Swift side had Rust's level of memory safety.

In time we hope that it will get easier and easier to make the Swift side safer and safer.
For now users are responsible for some kinds of memory safety.

crates/swift-integration-tests/src/async_function.rs
```diff
@@ -13,14 +13,19 @@ mod ffi {
...
             arg: Result<AsyncResultOpaqueRustType1, AsyncResultOpaqueRustType2>,
         ) -> Result<AsyncResultOpaqueRustType1, AsyncResultOpaqueRustType2>;
         async fn rust_async_func_return_result_null_opaque_rust(
             succeed: bool,
         ) -> Result<(), AsyncResultOpaqueRustType2>;
+
+        // TODO: this is broken because RustString is not Sendable.
+        // Work around making String and other opaque types Sendable is tracked
+        // here: https://github.com/chinedufn/swift-bridge/issues/150
+        // async fn rust_async_reflect_string(string: String) -> String;
```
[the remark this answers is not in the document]
colinmarc: Uncommenting this line makes it fail to compile on Swift 6, and I think changing core code in this PR would be out of scope, no?

colinmarc:
Regardless of how you want to proceed on the `RustStr: Sendable` stuff, that needs a fix before I can uncomment the line here. I think I can proceed on #307 without this, however, as long as I don't port the async tests over to the new harness yet.

debanjanbasu:
Any idea, when this would be merged 🤔?

PR #314, opened 2025-02-02
Add attribute for Swift ownership support

chinedufn:
Add attribute for Swift ownership support

This commit introduces a new attribute that can be used to enable
experimental support for Swift ownership.

The new `__experimental_swift_ownership` can be used on an opaque Rust
type.

```rust
#[swift_bridge::bridge]
mod foo {
    extern "Rust" {
        #[swift_bridge(__experimental_swift_ownership)]
        type SomeType;
    }
}
```

This new attribute currently does nothing.
In future commits, when `swift-bridge` sees that a type has the
experimental ownership attribute it will generate code that takes
advantage of Swift 6's ownership features.

For instance, we will use Swift's `~Copyable` protocol to ensure that
when Swift has ownership over an opaque Rust type it cannot copy that
handle, which should make attempts to use-after-free a compile-time
error instead of a runtime error.


crates/swift-bridge-ir/src/parse/parse_extern_mod.rs
```diff
@@ -805,158 +805,6 @@ mod tests {
         assert_eq!(parse_errors(tokens).len(), 0,);
     }
 
-    /// Verify that we can parse the `already_declared` attribute.
```
chinedufn, line 808: tests moved to another file

crates/swift-bridge-ir/src/parse/parse_extern_mod/opaque_type_attributes.rs
```diff
@@ -144,3 +153,139 @@ impl Deref for OpaqueTypeAllAttributes {
...
+                extern "Rust" {
+                    #[swift_bridge(__experimental_swift_ownership)]
+                    type SomeType;
+                }
+            }
+        };
+        let attribs = unwrap_opaque_type_attributes(tokens, "SomeType");
+
+        assert_eq!(attribs.experimental_swift_ownership, true);
+    }
```
chinedufn, line 276: This is the only new test case

PR #276, opened 2024-06-11
Support failable initializers

NiwakaDev:
Related to #235.

This PR supports failable initializers.

Something like:

```rust
// Rust side
#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        #[swift_bridge(Equatable)]
        type FailableInitType;

        #[swift_bridge(init)]
        fn new() -> Option<FailableInitType>;
    }
}
```

```swift
// Swift side
let failableInitType = FailableInitType()
if failableInitType == nil {
    ~
} else {
   ~
}
```

See: [Swift Documentation - Failable Initializers](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/initialization/#Failable-Initializers)

PR #272, opened 2024-04-28
Support Option<OpaqueSwiftType>

Bright-Shard:
This adds support for bridging `Option<OpaqueSwiftType>` in `extern "Rust"` functions. This fixes #268, and makes the following now possible:

```rs
#[swift_bridge::bridge]
mod ffi {
    extern "Swift" {
        type SomeSwiftType;
    }

    extern "Rust" {
        fn option_arg(arg: Option<SomeSwiftType>);
        fn returns_option() -> Option<SomeSwiftType>;
    }
}
```

PR #328, opened 2025-05-22
Support `Equatable` on Rust `Copy` types

pearfalse:
Allows automatic use of `==` on opaque (`extern "Rust"`) copyable types. The correct code was already being generated for the Rust crate and C header, but was not exposed in Swift code.

For the given bridge module declaration:
```rust
// Rust
#[swift_bridge::bridge]
mod ffi {
    extern "Rust" {
        #[swift_bridge(Copy(12), Equatable)]
        type SomeOpaqueBytes;

        ~
    }
}
```

Writing the following Swift is now possible:
```swift
// Swift

// `SomeOpaqueBytes` is a struct type and does not need to be heap-allocated
let x = SomeOpaqueBytes(initWith: ~)
let y = SomeOpaqueBytes(initSomeOtherWay: ~)

// With this PR, we can now compare them with `==`
if x == y { // <--
    ~
}
```


crates/swift-integration-tests/src/opaque_type_attributes/copy.rs
```diff
@@ -1,7 +1,7 @@
 #[swift_bridge::bridge]
 mod ffi {
     extern "Rust" {
-        #[swift_bridge(Copy(6))]
+        #[swift_bridge(Copy(6), Equatable)]
```
chinedufn, line 4: Please create a separate type such as `RustCopyAndEquatableType` so that we can remain confident that `Copy` works without `Equatable`.

crates/swift-bridge-ir/src/codegen/generate_swift/opaque_copy_type.rs
```diff
@@ -112,11 +112,36 @@ extension {ffi_repr_name}: SwiftBridgeGenericCopyTypeFfiRepr {{}}"#,
         )
     };
 
+    let ext_equatable = if ty.attributes.equatable {
```
chinedufn, line 115: Please add a codegen test.
One benefit is that since our codegen tests run much more quickly than the integration tests they allow us to more easily iterate/fix/change the codegen as we work.

Example: https://github.com/chinedufn/swift-bridge/blob/c15279c5a412fbc597e6831673b6bf850ddb4217/crates/swift-bridge-ir/src/codegen/codegen_tests/opaque_rust_type.rs#L148-L209

chinedufn (commented):
Looks good. Left some minor feedback.

Please update your PR body to include an example sketch of code that was impossible before this PR, but now possible.

You can find examples of these sketches in other PRs. For example: https://github.com/chinedufn/swift-bridge/pull/276 https://github.com/chinedufn/swift-bridge/pull/273 https://github.com/chinedufn/swift-bridge/pull/272

chinedufn:
Thanks!

PR #355, opened 2025-12-08
Add #[swift_bridge::bridged] attribute macro for structs outside bridge modules

Choochmeque:
Adds `#[swift_bridge::bridged]` attribute macro that generates FFI glue code for structs defined outside of bridge modules

## Example Usage
```rust
// rust
#[swift_bridge::bridged]
pub struct UserData {
    pub id: u32,
    pub active: bool,
}

#[swift_bridge::bridge]
mod ffi {
    #[swift_bridge(already_declared, swift_repr = "struct")]
    struct UserData;  // No fields needed - zero duplication!

    extern "Rust" {
        fn get_user() -> UserData;
    }
}
```

```swift
// swift
let user = get_user()
print("User ID: \(user.id), Active: \(user.active)")

// Create from Swift side
let newUser = UserData(id: 42, active: true)
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -61,7 +61,8 @@ jobs:
         with:
           toolchain: stable
 
-      - name: Set Xcode version 
+      # Xcode 16.2 provides Swift 5.9+ which is required for typed throws.
+      - name: Set Xcode version
         run: sudo xcode-select -s /Applications/Xcode_16.2.app
 
       - name: Add rust targets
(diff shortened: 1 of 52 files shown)


chinedufn:
To date the project has been against supporting inline annotations.

Reasons are explained here https://github.com/chinedufn/swift-bridge/blob/e527dc7454bebbfab537d842990bf3a1672e676c/book/src/bridge-module/why-a-bridge-module/README.md

I'll close this, but if you believe that the current policy is wrong, please feel free to explain.

PR #362, opened 2025-12-23
Add Swift 6 compatibility for async function code generation

Choochmeque:
Add Swift 6 strict concurrency compatibility for generated async function code
Add `@MainActor` to test classes using XCTContext.runActivity
Add CI matrix to test with both Swift 5.0 and Swift 6.0

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,6 +55,8 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
+        # NOTE: As explained in `README.md`, the minimum supported Swift version is currently `6.0`.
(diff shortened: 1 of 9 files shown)


Choochmeque:
Hi @chinedufn,
generated async swift code, was not compiling with swift 6. This patch should fix it. 
Could you please review it when you have time.
Also, could you please make a new release after merge this PR?

chinedufn (commented):
Looks good thanks. Just one minor piece of feedback about making it clear that if Swift 5 ever becomes too difficult to maintain we can remove it from the test matrix.

chinedufn:
Published as `0.1.59` https://crates.io/crates/swift-bridge/0.1.59

